PR #8, opened 2016-07-26
Eclipse configuration

tomasbjerre:


cdancy:
LGTM. Thanks @tomasbjerre!

I do all of my development work on mac/linux and specifically on intellij which is why most of this stuff was non existent.

PR #10, opened 2016-07-26
Avoiding star import

tomasbjerre:
* As it is forbidden in Checkstyle config. To make build pass.


cdancy:
LGTM and thanks again @tomasbjerre !

PR #12, opened 2016-07-27
Pull request comments API

tomasbjerre:
* Only GET, DELETE and POST method implemented.
 * Changing BitbucketFallbacks.getErrors() so that it throws the original throwable. It was lost when the original errors was not a JSON object. Then a new InvocationTargetException occured and was thrown, and the original throwable lost.
 * Using Java Method Invocation Builder on API:s to create invocation builders. Making client code cleaner.
 * Using AssertJ and changing some assertions to show the actual failure reason.

### DIFF
--- a/src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiMockTest.java
+++ b/src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiMockTest.java
@@ -17,6 +17,11 @@
 
 package com.cdancy.bitbucket.rest.features;
 
+import static org.testng.Assert.assertNotNull;
+import static org.testng.Assert.assertTrue;
+
+import org.testng.annotations.Test;
+
 import com.cdancy.bitbucket.rest.BitbucketApi;
 import com.cdancy.bitbucket.rest.BitbucketApiMetadata;
 import com.cdancy.bitbucket.rest.domain.repository.Repository;
@@ -24,10 +29,6 @@ import com.cdancy.bitbucket.rest.internal.BaseBitbucketMockTest;
 import com.cdancy.bitbucket.rest.options.CreateRepository;
 import com.squareup.okhttp.mockwebserver.MockResponse;
 import com.squareup.okhttp.mockwebserver.MockWebServer;
-import org.testng.annotations.Test;
-
-import static org.testng.Assert.assertNotNull;
-import static org.testng.Assert.assertTrue;
 
 /**
  * Mock tests for the {@link RepositoryApi} class.
@@ -85,7 +86,6 @@ public class RepositoryApiMockTest extends BaseBitbucketMockTest {
         try {
             String projectKey = "PRJ";
             String repoKey = "myrepo";
-            CreateRepository createRepository = CreateRepository.create(repoKey, true);
             Repository repository = api.get(projectKey, repoKey);
             assertNotNull(repository);
             assertTrue(repository.errors().size() == 0);
@@ -106,7 +106,6 @@ public class RepositoryApiMockTest extends BaseBitbucketMockTest {
         try {
             String projectKey = "PRJ";
             String repoKey = "notexist";
-            CreateRepository createRepository = CreateRepository.create(repoKey, true);
             Repository repository = api.get(projectKey, repoKey);
             assertNotNull(repository);
             assertTrue(repository.errors().size() == 1);
--- a/src/main/java/com/cdancy/bitbucket/rest/domain/branch/Type.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/domain/branch/Type.java
@@ -17,13 +17,9 @@
 
 package com.cdancy.bitbucket.rest.domain.branch;
 
-import com.cdancy.bitbucket.rest.error.Error;
-import com.google.auto.value.AutoValue;
-import com.google.common.collect.ImmutableList;
-import org.jclouds.javax.annotation.Nullable;
 import org.jclouds.json.SerializedNames;
 
-import java.util.List;
+import com.google.auto.value.AutoValue;
 
 @AutoValue
 public abstract class Type {
--- a/src/main/java/com/cdancy/bitbucket/rest/domain/repository/Repository.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/domain/repository/Repository.java
@@ -17,16 +17,16 @@
 
 package com.cdancy.bitbucket.rest.domain.repository;
 
-import com.cdancy.bitbucket.rest.domain.project.Project;
-import com.cdancy.bitbucket.rest.domain.pullrequest.Links;
-import com.cdancy.bitbucket.rest.domain.pullrequest.ProjectKey;
-import com.cdancy.bitbucket.rest.error.Error;
-import com.google.auto.value.AutoValue;
-import com.google.common.collect.ImmutableList;
+import java.util.List;
+
 import org.jclouds.javax.annotation.Nullable;
 import org.jclouds.json.SerializedNames;
 
-import java.util.List;
+import com.cdancy.bitbucket.rest.domain.project.Project;
+import com.cdancy.bitbucket.rest.domain.pullrequest.Links;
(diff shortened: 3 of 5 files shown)


tomasbjerre:
Just adding this PR to make sure it builds.

cdancy:
Looks to be a good size PR. Ping when you think you've got things fleshed out and ready for review.

tomasbjerre:
This is a bit ugly:
```
PagedResponse<PullRequestComments> actualResponse = api.get("PRJ", "repo", 1, "basic_branching/file.txt", 0, 10);
```

Hard to see what the parameters actually are.

If the project is built with Java 8 (still built with Java 7 compatibility) it can use [Java Method Invocation Builder](https://github.com/tomasbjerre/java-method-invocation-builder). Then builders are generated so that the above call can be:

```
PagedResponse<PullRequestComments> actualResponse = PullRequestCommentsGetBuilder
 .withProject("PRJ")
 .withRepo("repo")
 .withId(1)
 .withPath("basic_branching/file.txt")
 .withStart(0)
 .withLimit(10)
 .invoke(api);
```

Should I add it?

cdancy:
@tomasbjerre will this change the code at all or just add more generated source? I want to keep compatibility with how jclouds structures and expects projects to look and behave. However, if this is just an added feature that developers can use, and doesn't require any extra coding on anyone else adding features in the futures, than I would be Ok adding it in.

cdancy:
Also ... the smaller the PR the better. It looks like you are trying to implement the whole API in one go which will make things harder to review. If we could break the PR up into smaller more manageable chunks (e.g. 1 PR for adding a comment and all its tests, 1 PR for deleting a comment and all its tests) that would be best and allow us to move forward potentially quicker.

tomasbjerre:
Perhaps you want to review now? I only implemented GET.

I added a live test and it looks like it works. Bu I left it disabled. Some of the live tests are failing for me:
 * BranchApiLiveTest. testCreateBranch
 * PullRequestApiLiveTest. createGetPullRequest
 * TagApiLiveTest. testCreateTag

Perhaps cleaning them up can be done in its own PR.

tomasbjerre:
I also added ***changes*** to pull request API now. Added it as its own commit to make review easier.

This should be all functionality that I need. How fast can you make a release with this?

cdancy:
I want to take a step back here. There are changes all over the place and I'd like to try and keep the PR(s) as terse and to the point as possible. Here is how I'd like to proceed and break things up:

1.  PR for PullRequestCommentApi.get
2. PR for PullRequestCommentApi.post
3. PR for PullRequestCommentApi.delete
4. PR for any code cleanup we'd like to make happen
5. PR for adding your `java-method-invocation-builder` plugin

For each PR lets try to touch only those files which are relevant to the task at hand and save any "cleanup" or "code fixing" work for a separate PR.

tomasbjerre:
Can it be one pr with separate commits for these things?

Den 28 juli 2016 17:45 skrev "Christopher Dancy" <notifications@github.com>:

tomasbjerre:
I splitted the commit into several now, should be just as easy to review [the commits](https://github.com/cdancy/bitbucket-rest/pull/12/commits) as it is to review several PR:s.

cdancy:
@tomasbjerre Ok. I would however like to remove your plugin `java-method-invocation-builder` and do that in a separate PR. This is not a requirement for any of this work and is more an added benefit. It does not fall inline with the jclouds project standard and I'm trying to keep an apples-to-apples structure here as those guys are extremely helpful when diagnosing issues surrounding this library and adding this in will  only make potential problems that much harder. I don't mind looking over things later, just not in this PR.

Also, and when running specific integration tests, you [can edit this line](https://github.com/cdancy/bitbucket-rest/blob/master/build.gradle#L105) to say `include "**/PullRequestCommentsApiLiveTest.class"`. Assuming you have a bitbucket instance you can reach be sure to pass in your credentials `./gradlew integTest -PtestBitbucketEndpoint=<my-bitbucket-url> -PtestBitbucketCredential=username:password`

tomasbjerre:
The plugin is now in its own PR.

src/main/java/com/cdancy/bitbucket/rest/features/PullRequestApi.java
```diff
@@ -102,4 +102,13 @@ PullRequest reopen(@PathParam("project") String project,
                        @PathParam("repo") String repo,
                        @PathParam("pullRequestId") int pullRequestId,
                        @QueryParam("version") int version);
+
+    @Named("pull-request:changes")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Path("/{project}/repos/{repo}/pull-requests/{pullRequestId}/changes")
+    @Fallback(PullRequestOnError.class)
+    @GET
+    PagedResponse<Change> changes(@PathParam("project") String project,
```
cdancy: Lets split this out into its own PR as its not related to PullRequestCommentsApi. Also, and for future reference, put all domain objects under their relevant package/folder. The above domain objects `Path`, `Changes`, and `PagedResponse` I would expect to see within `com.cdancy.bitbucket.rest.domain.pullrequest`

src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiMockTest.java
```diff
@@ -88,7 +89,7 @@ public void testGetRepository() throws Exception {
             String repoKey = "myrepo";
             Repository repository = api.get(projectKey, repoKey);
             assertNotNull(repository);
-            assertTrue(repository.errors().size() == 0);
+            assertThat(repository.errors()).isEmpty();
```
cdancy: While things like this are fine they don't pertain to the PullRequest at hand. I'd say remove commit `5b792fb9bbcc30785bfd540931fd206244ce45aa` altogether.

src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiMockTest.java
```diff
@@ -106,7 +106,6 @@ public void testGetRepositoryNonExistent() throws Exception {
         try {
             String projectKey = "PRJ";
             String repoKey = "notexist";
-            CreateRepository createRepository = CreateRepository.create(repoKey, true);
```
cdancy: Again, while cleanup and code fixing is great, lets create a separate PR for this work and remove commit `b301b4888e920866065c17a5f30e6583a37d5a52` as it doesn't pertain to `PullRequestCommentsApi`.

src/main/java/com/cdancy/bitbucket/rest/features/PullRequestApi.java
tomasbjerre: ***PagedResponse*** is generic and lots of resources are paged.
***Path*** and ***Changes*** are both shared with the commits resource.

Do you still think they should be put under pullrequest?

src/main/java/com/cdancy/bitbucket/rest/fallbacks/BitbucketFallbacks.java
```diff
@@ -135,11 +123,27 @@ public static MergeStatus createMergeStatusFromErrors(List<Error> errors) {
...
-    public static List<Error> getErrors(String output) {
-        JsonElement element = parser.parse(output);
+    public static List<Error> getErrors(Throwable throwable) {
+        checkNotNull(throwable, "throwable");
+
+        JsonElement element;
+        try {
+            element = parser.parse(throwable.getMessage());
+        } catch (Exception e) {
+            // Happens when getMessage contains malformed JSON. Perhaps other
```
cdancy: Seems we are re-working this file because of potentially malformed json? Where does this pop within bitbucket? If the json is malformed, or non-existent, things will fail hard anyway as they should.

tomasbjerre: Yes it will fail, and it did fail before. But the exception was lost. The stacktrace and message that says exactly what was actually wrong was lost.

PR #13, opened 2016-07-29
Adding Java Method Invocation Builder

tomasbjerre:
- Makes client code cleaner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,3 @@
 language: java
-jdk: oraclejdk7
+jdk: oraclejdk8
 script: ./gradlew clean build mockTest
--- a/README.md
+++ b/README.md
@@ -13,11 +13,25 @@ java client, based on jclouds, to interact with Bitbucket's REST API.
 Client's can be built like so:
 
       BitbucketClient client = new BitbucketClient.Builder()
-      .endPoint("http://127.0.0.1:7990") // Optional. Defaults to http://127.0.0.1:7990
-      .credentials("admin:password") // Optional.
-      .build();
+       .endPoint("http://127.0.0.1:7990") // Optional. Defaults to http://127.0.0.1:7990
+       .credentials("admin:password") // Optional.
+       .build();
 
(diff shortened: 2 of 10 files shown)


tomasbjerre:
I updated one of the test cases, ***PullRequestApiMockTest*** to show the use case. So the idea is that when invoking a method, that has alot of parameters, its more readable to do that with a builder. So that you have names where the parameters are supplied and you dont mix Strings with each other in a long list of parameters.

Whats happening when the ***@GenerateMethodInvocationBuilder*** annotation is added to a type is that builders are generated for each method in the type.

The ***PullRequestApi*** has a ***get()*** and for that ***PullRequestApiGetBuilder*** is generated containing.

```
package com.cdancy.bitbucket.rest.features;

import com.cdancy.bitbucket.rest.domain.pullrequest.PullRequest;
import java.lang.String;

public final class PullRequestApiGetBuilder {
  private String project;

  private String repo;

  private int pullRequestId;

  private PullRequestApiGetBuilder() {
  }

  public PullRequestApiGetBuilder withProject(final String project) {
    this.project = project;
    return this;
  }

  public PullRequestApiGetBuilder withRepo(final String repo) {
    this.repo = repo;
    return this;
  }

  public PullRequestApiGetBuilder withPullRequestId(final int pullRequestId) {
    this.pullRequestId = pullRequestId;
    return this;
  }

  public static PullRequestApiGetBuilder get() {
    return new PullRequestApiGetBuilder();
  }

  public PullRequest invoke(final PullRequestApi instance) {
    return instance.get(project,repo,pullRequestId);
  }
}
```

tomasbjerre:
License is Apache. I dont know how Versioneye determines the license of a project.

https://github.com/tomasbjerre/java-method-invocation-builder/blob/master/LICENSE
https://github.com/tomasbjerre/java-method-invocation-builder-annotations/blob/master/LICENSE

cdancy:
@tomasbjerre think the license file in the projects root needs to be named `LICENSE.txt` for versioneye to find it.

PR #19, opened 2016-12-07
Paging api

Gawen-pjr:
Opening another PR. Seems I did a false maneuver integrating your comments...

Sorry for the lost remarks...

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 group = com.cdancy
-version = 0.0.10-pjr
+version = 0.0.10-SNAPSHOT
 
 artifactoryURL = http://127.0.0.1:8080/artifactory
 artifactoryUser = admin
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 group = com.cdancy
-version = 0.0.10
+version = 0.0.10-pjr
 
 artifactoryURL = http://127.0.0.1:8080/artifactory
 artifactoryUser = admin
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 group = com.cdancy
-version = 0.0.10-SNAPSHOT
+version = 0.0.11-SNAPSHOT
 
 artifactoryURL = http://127.0.0.1:8080/artifactory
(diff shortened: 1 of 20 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 group = com.cdancy
-version = 0.0.10-pjr
+version = 0.0.10-SNAPSHOT
```
Gawen-pjr: In fact, it should actually be `0.0.11-SNAPSHOT`, since version 0.0.10 is already released, and you targeted this PR to v0.0.11 milestone.

Gawen-pjr:
Added mock and live test as expected. Mock test pass with success. For live test, I wait on travis check

cdancy:
@Gawen-pjr live tests are not wired together on travis. This is more due to Bitbucket and standing up an instance requires a license (I use a dev license for local development when necessary otherwise I connect to an instance we use here at work). Either way, I'll give the tests a live run tomorrow morning and see how things fare. Thanks AOT!

cdancy:
Integration tests pass as expected. Thanks @Gawen-pjr !!!

Did you need a new release with this change?

Gawen-pjr:
Not yet. Still adding new features. I'll come back with a new PR soon ;-)

cdancy:
@Gawen-pjr very good and thanks! Have you previously worked with jclouds before? I'm assuming yes just wanted to be sure. We're a small group all things considered. I ask because I've started another project [api-processor](https://github.com/cdancy/api-processor) which is an attempt to factor out the core processing of jclouds (the part which turns `Api`'s, and their methods, into something invokable) into its own project which will then eventually be used here and in a few other projects I have that use jclouds core which is a bit heavy weight. Take a peek if you're interested as I could use the help and don't come across many devs in the wild who understand this code ;)

Gawen-pjr:
It was acually my first time with jclouds, but I'm quite used to play with meta-programming, so I had no problem to understand it's philosophy.

For your other project, I'll take a peek as you suggested. But I can't engage myself, as I don't know if I'll have enough time for it. I have my own business and customers to manage...

PR #25, opened 2016-12-14
Factorize code with ErrorsWrapper and Utils.nullToEmpty

Gawen-pjr:
Just propagated usage of ErrorsWrapper interface and Utils.nullToEmpty static method for all domain classes.

PR #26, opened 2016-12-14
Factorize Links attributes in domain classes with LinksHolder interface

Gawen-pjr:
Same work as in PR #25, but for "Links" attributes this time.

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/domain/comment/Comments.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/domain/comment/Comments.java
@@ -25,13 +25,14 @@ import org.jclouds.json.SerializedNames;
 
 import com.cdancy.bitbucket.rest.domain.common.Error;
 import com.cdancy.bitbucket.rest.domain.common.ErrorsHolder;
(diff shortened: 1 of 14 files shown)


cdancy:
@Gawen-pjr can you add mock tests (simple assertion will suffice) for those features that have links to ensure they are not null, or are null, when expected?

Gawen-pjr:
Mock test added

Gawen-pjr:
I have a question regarding the potential next PR : 

What is the purpose of package `com.cdancy.bitbucket.rest.options`, and why is this package not a sub-package of `com.cdancy.bitbucket.rest.domain` (as it defines model-related classes) ?

If I were you, I'd put these classes in their corresponding domain package (for example `CreateProject` in `com.cdancy.bitbucket.rest.domain.project`). What do you think about this?

cdancy:
@Gawen-pjr lets keep them where they are. I agree with your assessment but I'm following the [jclouds](https://github.com/jclouds/jclouds-labs) model so as to keep things inline with their standards as much as possible.

cdancy:
Live tests pass as expected. Thanks for the PR @Gawen-pjr !!!

PR #42, opened 2017-03-05
Enabling Eclipse project generation

tomasbjerre:


build.gradle
```diff
@@ -46,6 +47,29 @@ ext.compatibilityVersion = '1.7'
 sourceCompatibility = compatibilityVersion
 targetCompatibility = compatibilityVersion
 
+sourceSets {
+    main.java.srcDirs = ['src/main/java'
+                        ,'src/generated/java' ]
```
cdancy: I assume `src/generated/java` is an eclipse thing? Is the need for this to pickup the `AutoValue*` classes?

build.gradle
```diff
@@ -46,6 +47,29 @@ ext.compatibilityVersion = '1.7'
...
+ generatedDir.mkdirs()
+}
+
+clean.doFirst {
+ File generatedDir = file("${rootDir}/src/generated/java")
+ if (generatedDir.exists()) {
+  delete generatedDir
+ }
+ generatedDir.mkdirs()
+}
```
cdancy: This would then get generated for all users and use cases and put the `generated` folder under `src` which I don't think is something we need?

build.gradle
tomasbjerre: Yes the need is to pickup AutoValue* classes.

And Yes its needed for Eclipse.

build.gradle
tomasbjerre: It can be changed to any folder. Do you have another suggestion on folder name? I may be wrong here, bit this is something I have that is working for me =)

cdancy: I use netbeans and sometimes intellij. I often don't have a need to see these classes but when I do it's easy enough through both IDE's to look within the filesystem view.

Furthermore the AutoValue classes are already available within the gradle `build/classes/*` folders. Is that not enough?

cdancy: Might be there is a way through the gradle eclipse plugin to specify these folders as source folders and only pick up the AutoValue classes.

tomasbjerre:
Added a new commit now. Its actually all that is needed. If manually adding the generated source folder from eclipse. And thats fine by me =)

cdancy:
This works. Thanks @tomasbjerre !

PR #44, opened 2017-03-05
Show the actual error when initializing live test

tomasbjerre:


cdancy:
Would you want to give this a go? I've had it on my backlog to remove testng and just use assertj instead.

https://github.com/joel-costigliola/assertj-core/blob/master/src/main/scripts/convert-testng-assertions-to-assertj.sh

cdancy:
Closing in favor of #46

PR #45, opened 2017-03-05
Adding comments action on pull requests

tomasbjerre:


.gitignore
```diff
@@ -5,6 +5,7 @@ bin
 out 
 src/generated
 .gradle
+test-output
```
cdancy: What is this for?

tomasbjerre:
I think I got it when running TestNG from Eclipse.

.gitignore
tomasbjerre: I think I got it when running TestNG from Eclipse.

cdancy:
Bitbucket documentation is always either hit or miss. This is one of those cases [where it's a miss IMO.](https://developer.atlassian.com/static/rest/bitbucket-server/latest/bitbucket-rest.html#idm45888278617264). Can `path` be null? What is it suppose to represent?

tomasbjerre:
Right, yes it should be:
https://developer.atlassian.com/static/rest/bitbucket-server/latest/bitbucket-rest.html#idm45888278617264

Path cannot be null.

I forgot to add POST, can I add that to this PR?

cdancy:
@tomasbjerre I'll take this one but lets put `POST` functionality in a separate PR. 

Are you able to run live tests? If so, and as part of the subsequent `POST` PullRequest, can you write 2 live tests: one for posting a comment and one for listing all comments? I recognize that I still have not yet made it stupid-simple to run live tests against a system so feel free to change the parameters to point at something more local to you to ensure all works as expected.

tomasbjerre:
I just noticed that POST [is actually implemented](https://github.com/cdancy/bitbucket-rest/blob/master/src/main/java/com/cdancy/bitbucket/rest/features/CommentsApi.java#L67)! So wont be doing that.

cdancy:
Forgot about that ;) [I actually use the more simpler version](https://github.com/cdancy/bitbucket-rest/blob/master/src/main/java/com/cdancy/bitbucket/rest/features/CommentsApi.java#L56) which is probably why this slipped my mind.

PR #49, opened 2017-03-08
Add new api call for BranchPermission

j0nathan33:


cdancy:
@j0nathan33 thanks for the PR! First things first: looks like we've got 2 separate changes here: one for the BranchesApi and one for the PullRequestApi. Lets break this into 2 separate PR's to make things easier to review.

j0nathan33:
Ok, I will do

PR #50, opened 2017-03-08
Add new field in PullRequest Object

j0nathan33:


src/main/java/com/cdancy/bitbucket/rest/domain/pullrequest/Properties.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.json.SerializedNames;
+
+/**
+ * Created by jdoire on 07/03/2017.
+ */
+@AutoValue
+public abstract class Properties {
+    public abstract Long openTaskCount();
```
cdancy: If not nullable we can use `long` instead of boxed version. Same for `resolvedTaskCount`.

src/main/java/com/cdancy/bitbucket/rest/domain/pullrequest/Reference.java
```diff
@@ -31,11 +31,14 @@
...
+    public abstract String displayId();
+
     Reference() {
     }
 
-    @SerializedNames({ "id", "repository" })
-    public static Reference create(String id, MinimalRepository repository) {
-        return new AutoValue_Reference(id != null ? id : "refs/heads/master", repository);
+    @SerializedNames({"id", "repository", "displayId"})
+    public static Reference create(String id, MinimalRepository repository, String displayId) {
```
cdancy: Can we create overloaded constructor here. Previous can default to setting this property to say the `id`.

src/test/java/com/cdancy/bitbucket/rest/features/PullRequestApiLiveTest.java
```diff
@@ -48,10 +48,10 @@ public void createGetPullRequest() {
...
         MinimalRepository repository = MinimalRepository.create(repo, null, proj);
 
-        Reference fromRef = Reference.create("refs/heads/" + branchToMerge, repository);
-        Reference toRef = Reference.create(null, repository);
+        Reference fromRef = Reference.create("refs/heads/" + branchToMerge, repository, branchToMerge);
+        Reference toRef = Reference.create(null, repository, null);
         CreatePullRequest cpr = CreatePullRequest.create(randomChars, "Fix for issue " + randomChars, fromRef, toRef, null, null);
-        
+
         System.out.println("---------> CREATED PR: " + cpr);
```
cdancy: Mind taking out this line while you're here poking around? :)

cdancy:
@j0nathan33 thanks splitting up the PR. Added some comments to address. Looks good. Thanks.

src/main/java/com/cdancy/bitbucket/rest/domain/pullrequest/Reference.java
j0nathan33: I don't understand what you mean.

cdancy: Constructor was the wrong word, my apologies. Overloaded versions of `create`:

```
public static Reference create(String id, MinimalRepository repository) {

public static Reference create(String id, MinimalRepository repository, String displayId) {
  
```

The first can delegate to the second passing in whatever value `id` is to `displayId`.

src/main/java/com/cdancy/bitbucket/rest/domain/pullrequest/Properties.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.json.SerializedNames;
+
+/**
+ * Created by jdoire on 07/03/2017.
+ */
+@AutoValue
+public abstract class Properties {
+    public abstract long openTaskCount();
```
cdancy: Can we put a newline between class definition and `openTaskCount`?

src/main/java/com/cdancy/bitbucket/rest/domain/pullrequest/Properties.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package com.cdancy.bitbucket.rest.domain.pullrequest;
+
+import com.google.auto.value.AutoValue;
+import org.jclouds.json.SerializedNames;
+
+/**
+ * Created by jdoire on 07/03/2017.
+ */
```
cdancy: Lets remove these `Created by` comments.

src/main/java/com/cdancy/bitbucket/rest/domain/pullrequest/Reference.java
j0nathan33: Id and displayId isn't same value from api. Exemple for master branch : 
id: "refs/heads/master",
displayId : "master"

cdancy: Ok. Then maybe check for non-null and if so split on `/` and use last index? Something like this might work.

```
public static Reference create(String id, MinimalRepository repository) { 
    String displayId = null;
    if (id != null) {
        String parts [] = id.split("/");
        displayId = parts[parts.length];
    }
    return create(id, repository, displayId);
}
```

cdancy:
@j0nathan33 minor issue with the "overloaded" change but I'll fix that on this end. Thanks for the PR!

PR #52, opened 2017-03-09
Change 'getComment' endpoint to 'fileComments' and add live tests.

cdancy:


cdancy:
@tomasbjerre FYI: `comments` endpoint changed to `fileComments` to better align with what the actual call does.

PR #55, opened 2017-03-20
PullrequestApi : Add 2 new call (Activities & Participants)

j0nathan33:

PR #57, opened 2017-04-10
Add call to get List user in bitbucket by group

j0nathan33:
Need merge this pullRequest : https://github.com/cdancy/bitbucket-rest/pull/55 before this one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Can be sourced from jcenter like so:
 	<dependency>
 	  <groupId>com.cdancy</groupId>
 	  <artifactId>bitbucket-rest</artifactId>
-	  <version>0.0.13</version>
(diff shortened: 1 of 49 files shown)


cdancy:
Hey @j0nathan33. Just getting back from vacation and am a little behind on things. Quickly looking over this PR it seems it has the changes from #55 here? Is that necessary? Any particular reason why we should merge this PR first?

j0nathan33:
Merge first #55. For me, I need code in #55 to compile another internal application. If I don't include #55, It's impossible for me do some live test in bitbucket.

cdancy:
@j0nathan33 I'll merge this in but the way you created the `AdminApiLiveTest` is a little wonky. I'll fix that now as it's no big deal. Thanks again!

PR #60, opened 2017-05-08
Add new call for create repo permission

j0nathan33:


cdancy:
@j0nathan33 can you break this PR up into smaller more digestible chunks? I'm thinking 1 PR per adde endpoint which if I'm following your code correctly would get 4 different PR's. It's much easier to review code that is smaller and more manageable and allows me to test and get your changes in faster.

j0nathan33:
Ok, I will do it.

PR #62, opened 2017-05-08
Create/Delete permission group in repo

j0nathan33:


cdancy:
LGTM but will need some live tests before we can merge.

src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiLiveTest.java
```diff
@@ -119,6 +119,18 @@ public void testListPermissionByGroup() {
         assertThat(permissionsPage.values()).isEmpty();
     }
 
+    @Test(dependsOnMethods = "testGetRepository")
+    public void testCreatePermissionByGroupNonExistent() {
+        boolean success = api().createPermissionsByGroup(projectKey, repoKey, "REPO_WRITE", randomString());
```
cdancy: This actually returns a `204 no content`. Can we update the test to reflect as much as well as the mock test to set the correct return code?

cdancy:
One minor hiccup and I would like to see actual live tests that can succeed and work. Something like the below would suffice:

```
    @Test 
    public void testDeletePermissionByGroupNonExistent() {
        boolean success = api().deletePermissionsByGroup(projectKey, repoKey, "some-name");
        assertThat(success).isTrue();
    }
    
    @Test (dependsOnMethods = "testDeletePermissionByGroupNonExistent")
    public void testCreatePermissionByGroupNonExistent() {
        boolean success = api().createPermissionsByGroup(projectKey, repoKey, "REPO_WRITE", "some-name");
        assertThat(success).isTrue();
    }
```

src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiLiveTest.java
j0nathan33: I try with last bitbucket (5.0) and Bitbucket return this error : 

{
  "errors": [
    {
      "context": null,
      "message": "No such groups: 123",
      "exceptionName": null
    }
  ]
}

j0nathan33:
For me, default group in bitbucket is stash-users.

src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiLiveTest.java
j0nathan33: This test is good, because It try to create permission group with group don't existent.

cdancy:
> For me, default group in bitbucket is stash-users.

@j0nathan33 yeah that works. One of these days we'll have to get a proper integration test flow going instead of this ad-hoc process I have now. That would be an awesome story/ISSUE to take on ;)

cdancy:
In either event LGTM. Thanks @j0nathan33 !

PR #64, opened 2017-05-08
Create/delete permission users in repo

j0nathan33:


cdancy:
Nothing stick out as off here. Just rebase off of master, fix conflicts, and we should be good to go.

cdancy:
All tests pass as expected. Thanks again @j0nathan33 !

cdancy:
FYI: added some more live tests for this PR. It was easy enough to do so I just knocked it out while I was here.

PR #68, opened 2017-05-23
Add possibility to get branch Model Condition

j0nathan33:


src/main/java/com/cdancy/bitbucket/rest/options/CreateBranchModelConfiguration.java
```diff
@@ -0,0 +1,50 @@
+/*
```
cdancy: Should this file be here? Doesn't seem to be used

src/test/java/com/cdancy/bitbucket/rest/features/BranchApiLiveTest.java
```diff
@@ -136,12 +140,49 @@ public void testDeleteBranchPermission() {
...
+    }
+
     @AfterClass
     public void fin() {
         boolean success = api().updateDefault(projectKey, repoKey, defaultBranchId);
         assertThat(success).isTrue();
         success = api().delete(projectKey, repoKey, "refs/heads/" + branchName);
         assertThat(success).isTrue();
+        if (branchModelConfiguration != null) {
+            checkDefaultBranchConfiguration();
```
cdancy: Do we need to do this check twice?

src/test/java/com/cdancy/bitbucket/rest/features/BranchApiLiveTest.java
```diff
@@ -136,12 +140,49 @@ public void testDeleteBranchPermission() {
         }
     }
 
+    @Test(dependsOnMethods = {"testCreateBranch", "testListBranches"})
+    public void testGetBranchModelConfiguration() {
+        branchModelConfiguration = api().getModelConfiguration(projectKey, repoKey);
+        checkDefaultBranchConfiguration();
```
cdancy: If we don't need to do this check twice than lets remove the `checkDefaultBranchConfiguration` method and instead add that code here.

src/test/java/com/cdancy/bitbucket/rest/features/BranchApiLiveTest.java
```diff
@@ -17,17 +17,16 @@
```
cdancy: Another LiveTest to check for errors? Maybe named something like `testGetBranchModelConfigurationOnError()`?

src/test/java/com/cdancy/bitbucket/rest/features/BranchApiLiveTest.java
j0nathan33: Oops, This line is for update

src/main/java/com/cdancy/bitbucket/rest/options/CreateBranchModelConfiguration.java
j0nathan33: I use in another PR (#69).

src/test/java/com/cdancy/bitbucket/rest/features/BranchApiLiveTest.java
cdancy: Seen you pushed a commit but missed this. Not sure if you were planning on addressing with a subsequent commit.

j0nathan33: If I remove it, I need to readd it in another PR (#69)

cdancy: Gotcha ... yeah looking at that PR now. Ok lets keep it then.

src/test/java/com/cdancy/bitbucket/rest/features/BranchApiLiveTest.java
```diff
@@ -136,6 +140,48 @@ public void testDeleteBranchPermission() {
...
+        assertThat(configuration.errors()).isNotEmpty();
+        assertThat(configuration.development()).isNull();
+        assertThat(configuration.production()).isNull();
+        assertThat(configuration.types()).isEmpty();
+    }
+
+    private void checkDefaultBranchConfiguration() {
+        assertThat(branchModelConfiguration).isNotNull();
+        assertThat(branchModelConfiguration.errors().isEmpty()).isTrue();
+        assertThat(branchModelConfiguration.development().refId()).isNotNull();
```
cdancy: This was actually null when I ran the tests. Seeing as how annotated with @Nullable not sure if we should check for it here? I'm going to merge this now and remove the check. If you have a strong objection to this let me know and we can follow up on #69

PR #69, opened 2017-05-23
Add possibility to update branch Model Condition

j0nathan33:


cdancy:
Please rebase off of master to fix conflicts.

src/main/java/com/cdancy/bitbucket/rest/domain/branch/Type.java
```diff
@@ -17,24 +17,41 @@
...
+
+    public abstract TypeId id();
 
+    @Nullable
     public abstract String displayName();
 
     public abstract String prefix();
 
+    @Nullable
+    public abstract Boolean enabled();
```
cdancy: Can this really be Null? If not lets use unboxed boolean here and below in the `create` method.

j0nathan33: I check with api and It return some null value.

j0nathan33: I check with api and It return some null value.

cdancy:
All tests work as expected. Thanks again @j0nathan33 !

PR #84, opened 2017-05-29
Add possibility to delete Default Reviewers

j0nathan33:
My last PR for next few week :)


cdancy:
> My last PR for next few week :)

Don't ever feel bad about sending in code/PR's. The more the better ;)

Also ... It's going to be a few days until I can get to these. I'm fixing integration tests with this project so that they can all just run instead of the silly behavior we have now. Once that is done it will making testing changes much easier and reduce the PR turnaround time. Stay tuned.

j0nathan33:
It's not that. I am going on vacation

src/main/java/com/cdancy/bitbucket/rest/domain/branch/Matcher.java
```diff
@@ -25,8 +25,10 @@
 
     public enum MatcherId {
         RELEASE("RELEASE", "Release", "MODEL_CATEGORY", "Branching model category"),
```
cdancy: Think we should also switch `RELEASE` and `Release` around as well?

j0nathan33: For that, it's not possible to switch it because bitbucket have valid for that

cdancy:
Closing as this has been merged as part of #92

cdancy:
Thanks again @j0nathan33 !

PR #88, opened 2017-06-05
Add missing value in enum

j0nathan33:


cdancy:
@j0nathan33 pulling this in now. Thanks!

Issue #89, opened 2017-06-06
Integration Test does not working any more with last change on master


j0nathan33:
I run this command in my IDE : "gradlew integTest". After that, I check in the report and I have this error : 

> java.lang.AssertionError: 
> Expecting:
>  false
> to be equal to:
>  true
> but was not.
> 	at com.cdancy.bitbucket.rest.BaseBitbucketApiLiveTest.initGeneratedTestContents(BaseBitbucketApiLiveTest.java:186)
> 	at com.cdancy.bitbucket.rest.features.BranchApiLiveTest.init(BranchApiLiveTest.java:60)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.lang.reflect.Method.invoke(Method.java:498)
> 	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:108)
> 	at org.testng.internal.Invoker.invokeConfigurationMethod(Invoker.java:523)
> 	at org.testng.internal.Invoker.invokeConfigurations(Invoker.java:224)
> 	at org.testng.internal.Invoker.invokeConfigurations(Invoker.java:146)
> 	at org.testng.internal.TestMethodWorker.invokeBeforeClassMethods(TestMethodWorker.java:166)
> 	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:105)
> 	at org.testng.TestRunner.privateRun(TestRunner.java:744)
> 	at org.testng.TestRunner.run(TestRunner.java:602)
> 	at org.testng.SuiteRunner.runTest(SuiteRunner.java:380)
> 	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:375)
> 	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:340)
> 	at org.testng.SuiteRunner.run(SuiteRunner.java:289)
> 	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
> 	at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
> 	at org.testng.TestNG.runSuitesSequentially(TestNG.java:1301)
[...]

PR #90, opened 2017-06-06
Fix issues #89

j0nathan33:


cdancy:
Silly typo on my end. Thanks for fixing @j0nathan33 !

Issue #93, opened 2017-06-07
Junit Failing on DefaultReviewers when upgrade to Bitbucket 5.1


j0nathan33:
After upgrade my local setup to bitbucket 5.1, integration test don't working anymore for defaultReviewers. I will try to fix today or tomorrow before I going to vacancy

cdancy:
@j0nathan33 can you post the stacktrace here and/or what popped? Tests ran fine for me.

j0nathan33:
StackTrace : 

> java.lang.NullPointerException
> 	at com.cdancy.bitbucket.rest.features.DefaultReviewersApiLiveTest.validCondition(DefaultReviewersApiLiveTest.java:167)
> 	at com.cdancy.bitbucket.rest.features.DefaultReviewersApiLiveTest.testCreateCondition(DefaultReviewersApiLiveTest.java:69)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
> 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
> 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
> 	at java.lang.reflect.Method.invoke(Method.java:498)
> 	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:108)
> 	at org.testng.internal.Invoker.invokeMethod(Invoker.java:661)
> 	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:869)
> 	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1193)
> 	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:126)
> 	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:109)
> 	at org.testng.TestRunner.privateRun(TestRunner.java:744)
> 	at org.testng.TestRunner.run(TestRunner.java:602)
> 	at org.testng.SuiteRunner.runTest(SuiteRunner.java:380)
> 	at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:375)
> 	at org.testng.SuiteRunner.privateRun(SuiteRunner.java:340)
> 	at org.testng.SuiteRunner.run(SuiteRunner.java:289)
> 	at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
[...]

PR #94, opened 2017-06-07
Fix issues #93

j0nathan33:


cdancy:
Integration tests _almost_ pass as expected. I'll send in one minor fix to allow them to run on v4.x of bitbucket servers. Thanks again @j0nathan33 !

PR #99, opened 2017-06-27
Add convert for gson

j0nathan33:


build.gradle
```diff
@@ -32,16 +32,20 @@ repositories {
...
-    compileApi ('org.apache.jclouds:jclouds-core:2.0.1') 
-    compileApi ('com.google.auto.service:auto-service:1.0-rc3') 
+    compile ('com.google.code.gson:gson:2.5') {
+        force = true
+    }
+    compileApi ('org.apache.jclouds:jclouds-core:2.0.1')
+    compileApi ('com.google.auto.service:auto-service:1.0-rc3')
     compileApi ('com.google.auto.value:auto-value:1.4.1')
-    
+    compileApi ('com.ryanharter.auto.value:auto-value-gson:0.4.5')
```
j0nathan33: It's not last version because jcloud don't support gson 2.8 (https://issues.apache.org/jira/browse/JCLOUDS-1160)

build.gradle
```diff
@@ -32,16 +32,20 @@ repositories {
 }
 
 dependencies {
-    compileApi ('org.apache.jclouds:jclouds-core:2.0.1') 
-    compileApi ('com.google.auto.service:auto-service:1.0-rc3') 
+    compile ('com.google.code.gson:gson:2.5') {
+        force = true
```
j0nathan33: I force version of gson because jcloud don't support gson 2.8.0 (https://issues.apache.org/jira/browse/JCLOUDS-1160)

cdancy:
I was literally just looking at this project this morning :)

The only concern I have is surrounding empty collections which is noted below. I assume the default is to set them to null but we standardize on empty collections. Is that set somewhere here that I missed?

https://github.com/rharter/auto-value-gson#compiler-options

cdancy:
Furthermore ... it seems like we are adding a new method `typeAdapter` to all AutoValue classes but keeping the old `create` methods? I would think we would want to do one or the other but not sure if you can easily do so as this is part of jclouds core. Are these new methods even being used in this context?

cdancy:
And if you really want to do this, which in certainly in favor of, you might want to look at the GsonModule in jclouds and then we would probably have to provide our own here.

j0nathan33:
I need to add typeAdapter because my application try to save it in DB with json format.

1) I try to add 'autovaluegson.defaultCollectionsToEmpty' and each time, I try to add it, build fail but I haven't expert of gradle.
2) For typeAdapter and create methods, we need to keep both because jcloud don't support typeAdapter. 
3) I try GsonModule from jcloud and we need to edit all class to put errors in first arguments of create method because GsonModule do some weird association.

Exemple with errors at the end : 
{
    "errors": 0,
    "requiredAllApprovers": {
        "defaultStrategy": {
            "enabled": true,
            "id": "ff"
        },
        "strategies": [
            {
                "enabled": false,
                "id": "no-ff"
            },
            {
                "enabled": true,
                "id": "ff"
            },
            {
                "enabled": false,
                "id": "ff-only"
            },
            {
                "enabled": false,
                "id": "squash"
            },
            {
                "enabled": false,
                "id": "squash-ff-only"
            }
        ],
        "type": "REPOSITORY"
    },
    "requiredAllTasksComplete": false,
    "requiredApprovers": false,
    "requiredSuccessfulBuilds": 0
}

Exemple with error in first arguments : 
{
    "mergeConfig": {
        "defaultStrategy": {
            "enabled": true,
            "id": "ff"
        },
        "strategies": [
            {
                "enabled": false,
                "id": "no-ff"
            },
            {
                "enabled": true,
                "id": "ff"
            },
            {
                "enabled": false,
                "id": "ff-only"
            },
            {
                "enabled": false,
                "id": "squash"
            },
            {
                "enabled": false,
                "id": "squash-ff-only"
            }
        ],
        "type": "REPOSITORY"
    },
    "requiredAllApprovers": false,
    "requiredAllTasksComplete": false,
    "requiredApprovers": 0,
    "requiredSuccessfulBuilds": 0
}

cdancy:
Because this library is built on top of jclouds I'm not sure if we can even get away with this functionality as TypeAdapter is a requirement. Do you see some other way around this? What is the use-case here? Curiosity or just hacking or something else?

j0nathan33:
My use-case is, we have 300 repo in my local instance and I need to automate update/creation of repo in bitbucket. My application save template in Database with object convert to json.

If library don't support fonction TypeAdapter, The workarround is to modify fonction creator to put List<Error> in first parameter.

cdancy:
@j0nathan33 but can't that be done outside of this library? Seems like you're more/less saving the entire "metedata" of the database and then re-importing it elsewhere? 

Furthermore, if only you can take advantage of this, and subsequently puts the burden on devs to add yet another constructor of sorts to each domain object, that seems like an awful lot of code to maintain on our end for only your use-case.

j0nathan33:
I try to create new abstract class but I have same error "public com.cdancy.bitbucket.rest.domain.branch.Matcher() with no args".

I have 3 solution for my problem : 
1) Add function TypeAdapter in all domain
2) Modify parameter order for all function create to correspond to field order
3) Clone library to my local repo and add TypeAdapter for all domain.

cdancy:
@j0nathan33 what exactly are you trying to save to the database? Can you be more specific? With this change how do you plan on using it? 

Maybe we can come up with a better solution that exists outside of this scenario.

j0nathan33:
In DB, I try to save in json format some domain object. When I try to use GsonModule from jcloud, json converter don't working because create function don't have same order from field.

Exemple wrong : 
```
{
    "errors": 0,
    "requiredAllApprovers": {
        "defaultStrategy": {
            "enabled": true,
            "id": "ff"
        },
        "strategies": [
            {
                "enabled": false,
                "id": "no-ff"
            },
            {
                "enabled": true,
                "id": "ff"
            },
            {
                "enabled": false,
                "id": "ff-only"
            },
            {
                "enabled": false,
                "id": "squash"
            },
            {
                "enabled": false,
                "id": "squash-ff-only"
            }
        ],
        "type": "REPOSITORY"
    },
    "requiredAllTasksComplete": false,
    "requiredApprovers": false,
    "requiredSuccessfulBuilds": 0
}
```

If I edit create from PullRequestSettings for that : 
```
@SerializedNames({"errors", "mergeConfig", "requiredAllApprovers", "requiredAllTasksComplete", "requiredApprovers", "requiredSuccessfulBuilds"})
    public static PullRequestSettings create(@Nullable List<Error> errors, MergeConfig mergeConfig, Boolean requiredAllApprovers, Boolean requiredAllTasksComplete, Long requiredApprovers, Long requiredSuccessfulBuilds) {
        return new AutoValue_PullRequestSettings(Utils.nullToEmpty(errors), mergeConfig, requiredAllApprovers, requiredAllTasksComplete, requiredApprovers, requiredSuccessfulBuilds);
    }
```
GsonModule will convert objet to this : 
```
{
    "mergeConfig": {
        "defaultStrategy": {
            "enabled": true,
            "id": "ff"
        },
        "strategies": [
            {
                "enabled": false,
                "id": "no-ff"
            },
            {
                "enabled": true,
                "id": "ff"
            },
            {
                "enabled": false,
                "id": "ff-only"
            },
            {
                "enabled": false,
                "id": "squash"
            },
            {
                "enabled": false,
                "id": "squash-ff-only"
            }
        ],
        "type": "REPOSITORY"
    },
    "requiredAllApprovers": false,
    "requiredAllTasksComplete": false,
    "requiredApprovers": 0,
    "requiredSuccessfulBuilds": 0
}
```

j0nathan33:
Do you have any update for this PR ?

cdancy:
@j0nathan33 I don't think I can reasonably pull this in. It would require all users to add this additional block to their domain objects for no benefit to them. I have a side-project going on to move away from jclouds altogether and use a different driver in the background (though it will look and feel in exactly the same way) and this makes more sense to put it there but it's not quite at that stage yet.

cdancy:
Closing PR. While I liked the idea it's too much to ask for developers to code also for this for no benefit or requirement to them IMO. Will re-open in the future should I revisit this area of the code.

PR #141, opened 2018-05-02
Fix to account for Bitbucket not handing back a list of 'error' objects.

cdancy:
@martinda take a look at this PR and let me know what you think. Just whipped it up but it _should_ solve the problem. We will check if the returned json contains `errors` (i.e. meaning a list) then check for a single `error` and then fallback to the defaults.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,7 +108,7 @@ task integTest(type: Test, dependsOn: ['mockTest']) {
         }
     }
     useTestNG() 
-    include '**/BitbucketClientLiveTest.class'
+    include '**/**LiveTest.class'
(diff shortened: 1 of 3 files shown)


martinda:
This looks better than the fix I was working on. I have pulled down this PR and started to work with it. So far so good. Thank you!

cdancy:
@martinda np. Sorry if I stepped on your toes here but I was taking a look at this while in a meeting and it was bugging me as to what was actually going on.

martinda:
This looks good to me.

martinda:
Thank you!

Issue #142, opened 2018-05-07
Failed to parse Null description in build-status response


Joni-Lover:
Failed to parse Null description in build-status response

### Expected Behavior
When calling **buildStatusApi().status(commit, 1, 100)** and 'commit' has build-status with **null** description I should get correct response such as below:

`
[Status{dateAdded=1525262228566, description=null, key=0982e632-9d99-4f68-93f5-e50c3f6609c7, name=0982e632-9d99-4f68-93f5-e50c3f6609c7, state=FAILED, url=http://URL}]
`

### Current Behavior
The **null** description can't be parsed:

`
SEVERE: Error parsing input: Null description
java.lang.NullPointerException: Null description
	at com.cdancy.bitbucket.rest.domain.build.AutoValue_Status.<init>(AutoValue_Status.java:25)
	at com.cdancy.bitbucket.rest.domain.build.Status.create(Status.java:51)
`

### Context
I am testing for invalid user inputs with pre-creating build-status on 'commit' with null description.

### Steps to Reproduce (for bugs)
```groovy
@Grab(group='com.cdancy', module='bitbucket-rest', version='2.1.2')

import com.cdancy.bitbucket.rest.BitbucketClient

BitbucketClient client = BitbucketClient.builder().build()
println(client.buildStatusApi().status(commit, 1, 100).values())
```

### Your Environment
MacOX command line

cdancy:
Change has been merged to master. Thanks @Joni-Lover

Joni-Lover:
@cdancy Thank you very match !

PR #143, opened 2018-05-07
Fix nullable description in build-status

Joni-Lover:
SEVERE: Error parsing input: Null description

#142

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/domain/build/Status.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/domain/build/Status.java
@@ -19,6 +19,7 @@ package com.cdancy.bitbucket.rest.domain.build;
 
 import com.google.auto.value.AutoValue;
 import org.jclouds.json.SerializedNames;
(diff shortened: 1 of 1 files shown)


src/main/java/com/cdancy/bitbucket/rest/domain/build/Status.java
```diff
@@ -31,21 +32,23 @@
...
 
     @SerializedNames({"dateAdded", "description", "key", "name", "state", "url"})
-    public static Status create(final long dateAdded, 
-            final String description, 
-            final String key, 
-            final String name, 
+    public static Status create(final long dateAdded,
+            @Nullable final String description,
+            final String key,
+            @Nullable final String name,
```
cdancy: You can remove these 2 @Nullable annotations as it's only required on the variable itself for AutoValue to correctly handle things.

cdancy:
Pulling this in now and will massage the @Nullable's. Look out for a bugfix/patch release shortly.

PR #147, opened 2018-08-08
Fix for BranchApi.getDefault returning a 204/null and updated merge s…

cdancy:
@martinda care to give this branch a go and see how things fare?


martinda:
There is no live test for the missing branch case, I am trying to add one, but the `@BeforeClass` is tricky to bypass in `BranchApiLiveTest`. I need a test where nothing is pushed to the repo. For this, the only way it to spilt `initGeneratedTestContents` like I have done [here](https://github.com/martinda/bitbucket-rest/pull/1/commits/2af46275734f70d8b61887f14cab63b698c95b3e). Do you have suggestions for a better way to add such test?

Issue #146, opened 2018-08-08
Some errors running live tests


martinda:
I am not sure why, but I get failures running the live tests (master branch).

My question is: are live tests in a state where they are known to pass? If so, with which version of Bitbucket?

### Expected Behavior

Live tests should pass.

### Current Behavior

139 tests completed, 9 failed, 67 skipped

> Task :integTest FAILED

### Context

I am getting ready to add a couple new features (logging, create fork) but first I want to run the live tests.

### Steps to Reproduce (for bugs)

Running on a linux localhost:

```
./gradlew integTest \
    -Ptest.bitbucket.endpoint=http://localhost:7990 \
    -PtestBitbucketRestCredentials=admin:passwd
```

The executive summary of the problem I see is that Bitbucket sometimes returns HTTP 204 meaning no-content, but bitbucket-rest appears to expect content, and throws an exception.

An example of this HTTP 204 error is:
```
23:08:41.774 [Test worker] DEBUG org.jclouds.rest.internal.InvokeHttpMethod - >> invoking project:get
23:08:41.775 [Test worker] DEBUG org.jclouds.http.internal.JavaUrlHttpCommandExecutorService - Sending request -2035848655: GET http://127.0.0.1:7990/rest/api/1.0/projects/DEV HTTP/1.1
23:08:41.775 [Test worker] DEBUG jclouds.headers - >> GET http://127.0.0.1:7990/rest/api/1.0/projects/DEV HTTP/1.1
23:08:41.775 [Test worker] DEBUG jclouds.headers - >> Accept: application/json
23:08:41.775 [Test worker] DEBUG jclouds.headers - >> Authorization: Basic bWFydGluOm1hcnRpbg==
23:08:41.805 [Test worker] DEBUG org.jclouds.http.internal.JavaUrlHttpCommandExecutorService - Receiving response -2035848655: HTTP/1.1 200 null
23:08:41.805 [Test worker] DEBUG jclouds.headers - << HTTP/1.1 200 null
23:08:41.805 [Test worker] DEBUG jclouds.headers - << Transfer-Encoding: chunked
23:08:41.805 [Test worker] DEBUG jclouds.headers - << X-AREQUESTID: @12IZ2M8x1388x708x0
23:08:41.805 [Test worker] DEBUG jclouds.headers - << X-ASEN: SEN-11797004
23:08:41.805 [Test worker] DEBUG jclouds.headers - << X-Content-Type-Options: nosniff
[...]

PR #149, opened 2018-08-10
Grant predefined test user the repo read permission in the PullRequestApiLiveTest

martinda:
This allows the PullRequestApiLiveTest to pass without relying on external test setup.

As [discussed](https://github.com/cdancy/bitbucket-rest/issues/146#issuecomment-411485691) in #146 this fixes the problem related to the test user not having the right permissions.

CC @cdancy


cdancy:
Much better and LGTM

Issue #150, opened 2018-08-10
TagApiLiveTest fails with HTTP 405


martinda:
### Expected Behavior

The live test for deleting a tag should pass.

### Current Behavior

The live test for deleting a tag fails:

```
20:58:35.279 [Test worker] DEBUG org.jclouds.rest.internal.InvokeHttpMethod - >> invoking tag:delete
20:58:35.279 [Test worker] DEBUG org.jclouds.http.internal.JavaUrlHttpCommandExecutorService - Sending request -994819979: DELETE http://127.0.0.1:7990/rest/api/1.0/projects/DEV/repos/BEJQTXJXED/tags/HAMTYOHEUL HTTP/1.1
20:58:35.280 [Test worker] DEBUG jclouds.headers - >> DELETE http://127.0.0.1:7990/rest/api/1.0/projects/DEV/repos/BEJQTXJXED/tags/HAMTYOHEUL HTTP/1.1
20:58:35.280 [Test worker] DEBUG jclouds.headers - >> Accept: application/json
20:58:35.280 [Test worker] DEBUG jclouds.headers - >> Authorization: Basic bWFydGluOm1hcnRpbg==
20:58:35.288 [Test worker] DEBUG org.jclouds.http.internal.JavaUrlHttpCommandExecutorService - Receiving response -994819979: HTTP/1.1 405 null
20:58:35.288 [Test worker] DEBUG jclouds.headers - << HTTP/1.1 405 null
20:58:35.290 [Test worker] DEBUG jclouds.headers - << Transfer-Encoding: chunked
20:58:35.290 [Test worker] DEBUG jclouds.headers - << X-AREQUESTID: @1YRPAXLx1258x4323x0
20:58:35.290 [Test worker] DEBUG jclouds.headers - << X-ASEN: SEN-11797004
20:58:35.291 [Test worker] DEBUG jclouds.headers - << X-Content-Type-Options: nosniff
20:58:35.291 [Test worker] DEBUG jclouds.headers - << X-AUSERNAME: martin
20:58:35.291 [Test worker] DEBUG jclouds.headers - << Vary: Accept-Encoding
20:58:35.291 [Test worker] DEBUG jclouds.headers - << X-AUSERID: 1
20:58:35.291 [Test worker] DEBUG jclouds.headers - << Date: Fri, 10 Aug 2018 00:58:35 GMT
[...]

PR #151, opened 2018-08-10
Fix the delete Tag REST api path

martinda:
This fixes issue #150 

The fix is not elegant, but the underlying implementation by Atlassian is not elegant either. The merit this fix has is that it keeps all the TagApi operations under the same class.

If you have a suggestion on how to do this better, let me know.

CC @cdancy


cdancy:
Thanks again @martinda !

PR #148, opened 2018-08-09
Feature/live test for create branch in empty repo

martinda:
This builds on PR #147 by adding a live test that creates a branch in an empty repository.

PR #152, opened 2018-08-10
Remove the git test directories when live tests are done

martinda:
This addresses the concern raised in PR #148 in this [comment](https://github.com/cdancy/bitbucket-rest/pull/148#issuecomment-412190688).

The temporary git repositories created for the live test are now deleted when they are no longer needed.

### DIFF
--- a/src/test/java/com/cdancy/bitbucket/rest/TestUtilities.java
+++ b/src/test/java/com/cdancy/bitbucket/rest/TestUtilities.java
@@ -40,6 +40,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
 import java.util.UUID;
+import org.apache.commons.io.FileUtils;
 import org.jclouds.util.Strings2;
 
 /**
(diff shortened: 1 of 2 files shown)


src/test/java/com/cdancy/bitbucket/rest/TestUtilities.java
```diff
@@ -303,6 +303,28 @@ public static synchronized void terminateGeneratedTestContents(final BitbucketAp
...
+            File[] files = path.listFiles();
+            for (int i = 0; i < files.length; i++) {
+                if (files[i].isDirectory()) {
+                    deleteDirectory(files[i]);
+                } else {
+                    files[i].delete();
+                }
+            }
+        }
+        return (path.delete());
```
cdancy: This is good but if you wanted to add apache-io lib to the test dependencies I wouldn't mind that and you could this with a single call like: `Files.deleteDirectory(some-dir)`. I'm fine with this though but it would be one less thing we have to maintain.

martinda: Sure thing.

cdancy:
Taking this in now. Thanks for addressing this @martinda .

PR #168, opened 2019-01-15
If a task contains both a comment and a like then parsing would fail.

### LINKED ISSUE #167: Parsing error when a comment contains both a task _and_ a like.
**PR forthcoming** and will update issue once it is submitted. 

### Expected Behavior

Parsing of comment (for example, from the activities page) should return activity page objects to client code.

### Current Behavior

Parsing fails in bitbucket-rest prior to reaching client code. The error looks as follows:

```
    ERROR [2019-01-15 00:21:03,721] org.jclouds.http.functions.ParseJson: Error parsing input: java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 141 column 29 path $.values[4].comment.tasks[0].anchor.properties.
    ! java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 141 column 29 path $.values[4].comment.tasks[0].anchor.properties.
    ! at com.google.gson.stream.JsonReader.nextString(JsonReader.java:838)
    ! at com.google.gson.internal.bind.TypeAdapters$16.read(TypeAdapters.java:422)
```

Rest of stack trace omitted.

### Context

I can't parse PRs with liked comments; this is a problem for our development workflow.

1. Make a comment
2. Add a task
3. Like the comment

ftclausen:
The presence of a like was unexpected and would lead to a parsing failure:

```
    ERROR [2019-01-15 00:21:03,721] org.jclouds.http.functions.ParseJson: Error parsing input: java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 141 column 29 path $.values[4].comment.tasks[0].anchor.properties.
    ! java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 141 column 29 path $.values[4].comment.tasks[0].anchor.properties.
    ! at com.google.gson.stream.JsonReader.nextString(JsonReader.java:838)
    ! at com.google.gson.internal.bind.TypeAdapters$16.read(TypeAdapters.java:422)
// rest of stack omitted
```

Using [this diff](https://github.com/cdancy/bitbucket-rest/pull/106/files) as inspiration I have come up with a fix that can be used as is or as a base for something else.

`closes #167 `

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/domain/comment/TaskAnchor.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/domain/comment/TaskAnchor.java
@@ -21,13 +21,15 @@ import com.cdancy.bitbucket.rest.domain.pullrequest.Author;
 import com.cdancy.bitbucket.rest.BitbucketUtils;
 import com.google.auto.value.AutoValue;
 import java.util.Map;
+
+import com.google.gson.JsonElement;
 import org.jclouds.javax.annotation.Nullable;
 import org.jclouds.json.SerializedNames;
 
 @AutoValue
 public abstract class TaskAnchor {
 
-    public abstract Map<String, String> properties();
+    public abstract Map<String, JsonElement> properties();
 
     public abstract int id();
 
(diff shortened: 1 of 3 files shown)


cdancy:
Yeah we've had to do this in a few other places as well and is an acceptable solution when the returned json is not standard. LGTM.

Issue #169, opened 2019-02-13
Is there a possibility to retrieve the file path and line information for outdated comments?


stoesselt:
I know 2 possibilities to retrieve comments via this lib:
1. `commentsApi().fileComments(project, repo, pullRequestId, path, nextPageStart, null)` and
1. `pullRequestApi().listActivities(project, repo, pullRequestId, null, nextPageStart)` where I filter `action == COMMENTED`

In the Bitbucket PR overview page, a comment can be active or outdated. When it is outdated, a little symbol `OUTDATED` is shown in the top right corner of a comment.

The first method returns active comments on files. In these, the anchor property is set and I can state the file and line for the comment. **My first question:** Do you use the `anchorState` flag `ALL` to retrieve also orphaned comments (see [Bitbucket Rest API](https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-rest.html?utm_source=%2Fstatic%2Frest%2Fbitbucket-server%2Flatest%2Fbitbucket-rest.html&utm_medium=301#idm8286353872))?

The second method returns all comments done during the pull request. There are the ones from the first method and the outdated comments. These have no anchor (file and line) properties set. **Second question:** Does this library or the Bitbucket Rest API in general allow to fetch file and line information in another way?

I have not yet tried, but Bitbucket Rest API 2.0 seems to have suitable fields in the return data, see [inline part of return data](https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Busername%7D/%7Brepo_slug%7D/pullrequests/%7Bpull_request_id%7D/comments#get). **Third question:** Are there plans to support the new API version?

stoesselt:
As some further investigation turned out, I could achieve my desired result with only small changes to the library, see the pull request #171. So from my side, this issue can be closed.

cdancy:
@stoesselt thanks for the issue and PR. Apologies for getting around to this so late just been busy with other things at work and I'm the only maintainer here at the moment :)

Looking over your PR now...

cdancy:
Version `2.4.1` has been released. Thanks for the PR @stoesselt !!!

PR #171, opened 2019-02-14
Extended comments API to also fetch orphaned/outdated pull request co…

stoesselt:
This pull request is related to #169 .

It would be great, if you could release a new version of the lib in the next 2 weeks.


cdancy:
PR looks OK. Added the old method signature back with a Deprecated annotation so that this wont break existing users. If all goes well I will spin a point release today. Stay tuned...

cdancy:
Version `2.4.1` has been released with fix. Let me know how things go.

stoesselt:
@cdancy , fix works, thank you very much for your quick integration.

Issue #177, opened 2019-03-03
Write mock-tests for WebHookApi


cdancy:
Live tests are already written just need corresponding mock tests.

cdancy:
Resolved via #178

PR #178, opened 2019-03-05
Implementation of  mock-tests for WebHookApi

Joni-Lover:
Mock-tests for WebHookApi #177


cdancy:
@Joni-Lover thanks a ton for knocking these out! Looks like there is a pop on checkstyle. This always bites me in the @$$ when developing locally. If you want to catch them just run: `./gradlew clean build pTML` and that should trigger the checkstyle run.

cdancy:
LGTM. Thanks again for knocking this out @Joni-Lover !

PR #180, opened 2019-03-16
Issue-179 Implement last-modified API

adam-tylr:
Note the API [documentation](https://docs.atlassian.com/bitbucket-server/rest/6.0.0/bitbucket-rest.html#idp231) reads as if the "at" query parameter is optional. It says it will use the default branch. However, when writing integration tests, I always got an `com.atlassian.bitbucket.validation.ArgumentValidationException` when I excluded it saying "Streaming modifications requires a starting commit." so I made it required.

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/domain/file/LastModifiedSummary.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/domain/file/LastModifiedSummary.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cdancy.bitbucket.rest.domain.file;
+
+import com.cdancy.bitbucket.rest.BitbucketUtils;
+import com.cdancy.bitbucket.rest.domain.commit.Commit;
+import com.cdancy.bitbucket.rest.domain.common.Error;
+import com.cdancy.bitbucket.rest.domain.common.ErrorsHolder;
+import com.google.auto.value.AutoValue;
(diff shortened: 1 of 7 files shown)


src/main/java/com/cdancy/bitbucket/rest/domain/file/LastModifiedSummary.java
```diff
@@ -0,0 +1,52 @@
...
+
+    LastModifiedSummary() {
+    }
+
+    @SerializedNames({ "files", "latestCommit", "errors" })
+    public static LastModifiedSummary create(final Map<String, Commit> files,
+                 final Commit latestCommit,
+                 final List<Error> errors) {
+        return new AutoValue_LastModifiedSummary(BitbucketUtils.nullToEmpty(errors),
+            files,
```
cdancy: Lets all use `BitbucketUtils.nullToEmpty(files)` here as well.

src/main/java/com/cdancy/bitbucket/rest/domain/file/LastModifiedSummary.java
```diff
@@ -0,0 +1,52 @@
...
+import com.cdancy.bitbucket.rest.domain.common.ErrorsHolder;
+import com.google.auto.value.AutoValue;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.json.SerializedNames;
+
+import java.util.List;
+import java.util.Map;
+
+@AutoValue
+public abstract class LastModifiedSummary implements ErrorsHolder {
```
cdancy: Lets change the name to just `LastModified` to match the endpoint being used. The name here isn't horrible, it's actually fine, I just like to try and match the returning object with the endpoint name _if_ possible.

cdancy:
@adam-tylr couple minor nitpicky things to address but all-in-all the code is clean and there is not much to pick at. Good job and thanks.

adam-tylr:
@cdancy I agree with both of those. I even had the nulltoEmpty(files) then talked myself out of it.

cdancy:
@adam-tylr everything looks good on my end and thanks for the PR! I'm about to step out for a few hours but if you're looking for a point release with this change I can knock that out when I come back. Let me know.

cdancy:
I just merged this PR but noticed there were a few more things I wanted to tackle. I can do them when I get back or if you're up for it have it but here they are:

1. The fallback name should be changed to reflect the name changes above.
2. Change the endpoint/method-names from `listLastModified` to just `lastModified`

adam-tylr:
I'll knock those out real quick. You don't need to create a release yet. I noticed the list-files api is only half implemented. You can't specify a path to list them at. I might tackle that this weekend if I have time.

cdancy:
Another thought: I think i could probably write an HttpRequestFilter for this method that would allow us to combine both methods into a single version and if the path param is null we just scrub it from the request line. Thoughts?

adam-tylr:
I think that makes sense. Especially since there are several endpoints all set up the same way.

PR #181, opened 2019-03-16
Rename last modified API and fallback

adam-tylr:


cdancy:
@adam-tylr just pushed a commit for the "scrub null from request line" enhancement I talked about before. Let me know what you think and if it still reads clearly enough. I like the optimization being done here just want to be sure for end-users it makes sense and is not too convoluted.

cdancy:
Did also notice there were checkstyle/pmd violations on `FileApiMockTest` now that we should probably take care of.

adam-tylr:
I didn't know what pmd was until now but I have those all fixed. I think the scrub null will make sense to the end user. Even if they're unsure, they can send null, empty string, or "/" and it'll work properly. I added tests for those cases as well.

cdancy:
Awesome. If everything is OK with you I can go ahead and merge.

adam-tylr:
Yup, sounds good

cdancy:
Thanks for the PR. Let me know if you want a point release with this change in place. Have no problem knocking one out.

PR #188, opened 2019-03-21
Allow updating a task's status to RESOLVED or OPEN

### LINKED ISSUE #175: Discussion: Marking existing "RESOLVED" task as "OPEN".
I am looking into contributing functionality to mark an existing task as "OPEN" when it's already checked (i.e. "RESOLVED" in the API). I found the [Task](https://github.com/cdancy/bitbucket-rest/blob/master/src/main/java/com/cdancy/bitbucket/rest/domain/comment/Task.java) class that, as best I can figure out, is serialised and eventually passed onto Bitbucket with the correct values. I think the serialisation + massaging field names happens in [TaskApi](https://github.com/cdancy/bitbucket-rest/blob/master/src/main/java/com/cdancy/bitbucket/rest/features/TasksApi.java) class.

What would be ideal would be to have some setter method that can PUT the new task's status to the Bitbucket API. Given everything is being generated via Google AutoValue I'm not sure how to go about this. Just add a concrete setter to the existing AutoValue abstact Task class? That setter could directly send the appropriate JSON payload to change  the task status. However that does not feel in keeping with how AutoValue is used in this class.

ftclausen:
closes #175 

This is my first take at allowing a task's status to be changed after it has been made base on our discussions in the issue above. Not sure if it is what you had in mind; we can definitely change it as required.

Some notes

- I created an _UpdateTask_ class instead of _CreateTaskUpdate_ because it seemed a less confusing name
- Added mock tests to open task, resolve task and testing the error path
- Added live test (against local Bitbucket instance using dev license). This is just one test that both opens and resolves a task. Seems easier than messing with test dependencies.
- I could not get the live tests working with HTTP endpoints using _just_ a username so modified _TestUtilities_ to send the full user:pass string
- Also modified _TestUtilities_ to do a case insensitive comparison when trying to find the Bitbucket user

I'll comment some more inline.


src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
```diff
@@ -49,7 +50,12 @@
     private final String taskComment = TestUtilities.randomString();
     private int commentId = -1;
     private int taskId = -1;
-    
+    private int pullRequestId = -1;
+    private int repositoryId = -1;
+
+    private static final String TASK_OPEN = "OPEN";
```
ftclausen: I might just make these strings in the test - no need for a constant that's only used once.

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
```diff
@@ -121,15 +129,34 @@ public void testGetTaskOnError() {
...
+        final TaskAnchor anchor = instanceAlreadyOpen.anchor();
+
+        final UpdateTask updateTaskAlreadyOpen = UpdateTask.update(anchor, instanceAlreadyOpen.id(), TASK_RESOLVED, pullRequestId, repositoryId);
+        final Task instanceNowResolved = api().update(updateTaskAlreadyOpen.id(), updateTaskAlreadyOpen);
+        assertThat(instanceNowResolved).isNotNull();
+        assertThat(instanceNowResolved.errors().isEmpty()).isTrue();
+        assertThat(instanceNowResolved.state()).isEqualTo("RESOLVED");
+
+        final UpdateTask updateTaskBackToOpen = UpdateTask.update(anchor, instanceNowResolved.id(), TASK_OPEN, pullRequestId, repositoryId);
+        final Task instanceReopened = api().update(updateTaskBackToOpen.id(), updateTaskBackToOpen);
```
ftclausen: The `.update` method signature is a bit weird with passing `updateTaskBackToOpen.id()` _and_ the object `updateTaskBackToOpen` - normally I'd just pass the object. But this is required because of how `.update` maps to the Bitbucket API (out of our control). Maybe a wrapper method could be made instead but that was not in keeping with how the rest of the API is being done.

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiMockTest.java
```diff
@@ -97,11 +105,96 @@ public void testGetTask() throws Exception {
...
+            final TaskAnchor taskAnchor = api.get(taskId).anchor();
+            final UpdateTask updateTaskOpen = UpdateTask.update(taskAnchor, taskId, taskOpen, 1, 1);
+
+            final Task instanceNowOpen = api.update(updateTaskOpen.id(), updateTaskOpen);
+            assertThat(instanceNowOpen).isNotNull();
+            assertThat(instanceNowOpen.errors().isEmpty()).isFalse();
+            assertThat(instanceNowOpen.anchor()).isNull();
+            assertThat(instanceNowOpen.state()).isNull();
+
+            server.takeRequest();
```
ftclausen: I needed to access the second _RecordedRequest_ so just throwing away the first one above because it is already being tested elsewhere. There did not seem to be a way to access the second one directly and I did not want to make a test class wide _Task_ field because that breaks down test isolation. This seemed the lesser of two evils.

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiMockTest.java
```diff
@@ -97,11 +105,96 @@ public void testGetTask() throws Exception {
...
+        try {
+            final TaskAnchor taskAnchor = api.get(taskId).anchor();
+            final UpdateTask updateTaskResolved = UpdateTask.update(taskAnchor, taskId, taskResolved, 1, 1);
+
+            final Task instanceNowResolved = api.update(updateTaskResolved.id(), updateTaskResolved);
+            assertThat(instanceNowResolved).isNotNull();
+            assertThat(instanceNowResolved.errors().isEmpty()).isTrue();
+            assertThat(instanceNowResolved.state()).isEqualTo(taskResolved);
+
+            server.takeRequest();
```
ftclausen: See comment https://github.com/cdancy/bitbucket-rest/pull/188/files#r267995828

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
cdancy: Agreed.

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiMockTest.java
cdancy: Fair enough.

src/test/java/com/cdancy/bitbucket/rest/features/CommentsApiLiveTest.java
```diff
@@ -81,7 +81,7 @@ public void init() {
         
         final String randomChars = TestUtilities.randomString();
         final ProjectKey proj = ProjectKey.create(project);
-        final MinimalRepository repository = MinimalRepository.create(repo, null, proj);
+        final MinimalRepository repository = MinimalRepository.create(42, repo, null, proj);
```
cdancy: The big thing is: do the comments live tests still pass and work as expected though we've changed this POJO? I think it might be better to keep the original `create` method and just provide an overloaded one for the new use-case which is allowed. Or maybe I'm over thinking things...

src/test/java/com/cdancy/bitbucket/rest/features/PullRequestApiLiveTest.java
```diff
@@ -99,7 +99,7 @@ public void init() {
     private void createPullRequest() {
         final String randomChars = TestUtilities.randomString();
         final ProjectKey proj = ProjectKey.create(project);
-        final MinimalRepository repository = MinimalRepository.create(repo, null, proj);
+        final MinimalRepository repository = MinimalRepository.create(42, repo, null, proj);
```
cdancy: Why the number `42` everywhere? Meaning of life?

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
cdancy: Agreed.

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
cdancy: Yeah it is a bit weird and hate when api's do this type of thing. We _could_ simplify things and have it work as expected just not sure if making this more complicated is a good idea. Let me think about things overnight and get back to you tomorrow.

cdancy: Actually ... what happens if you live either one out? I suspect the pathParam is required but what happens if you leave things out of the object?

cdancy:
> I created an UpdateTask class instead of CreateTaskUpdate because it seemed a less confusing name
 FWIW: I only kept this naming convention because that's what jclouds uses and I wanted any folks familiar with that framework to be able to jump in here and hit the ground running. I personally don't care for the convention at all just did it for standards sake. With that in mind: I'm OK with this name :)

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
ftclausen: You mean leave out the object completely? So, for TaskApi, make it 

```
Task update(@PathParam("taskId") int taskId);
```

src/test/java/com/cdancy/bitbucket/rest/features/CommentsApiLiveTest.java
ftclausen: I routinely ran the live and mock tests while working on this and nothing failed. I'll look into overloading in any case

src/test/java/com/cdancy/bitbucket/rest/features/PullRequestApiLiveTest.java
ftclausen: It's the [ultimate answer to life, the universe, and everything](https://en.wikipedia.org/wiki/42_(number)#The_Hitchhiker's_Guide_to_the_Galaxy). I could make it random though?

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
cdancy: So when I look at the documentation for this task the body of the response can only be:
```
{
    "id": 99,
    "text": "Fix the missing and obsolete imports"
}
```
It looks like you're adding in a lot more properties with the `UpdateTask` class.

src/main/java/com/cdancy/bitbucket/rest/options/UpdateTask.java
```diff
@@ -0,0 +1,49 @@
...
+    public abstract int pullRequestId();
+
+    public abstract int repositoryId();
+
+
+    UpdateTask() {
+    }
+
+    @SerializedNames({"anchor", "id", "state", "pullRequestId", "repositoryId" })
+    public static UpdateTask update(final TaskAnchor anchor,
```
cdancy: lets change the method name to `create`. I'm fine with the class name being different however.

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
ftclausen: Good point, I have removed the pull request and repository ID properties.

src/test/java/com/cdancy/bitbucket/rest/features/CommentsApiLiveTest.java
ftclausen: I'm in the process of reverting this change after having simplified the UpdateTask properties

src/main/java/com/cdancy/bitbucket/rest/options/UpdateTask.java
ftclausen: Will do

src/test/java/com/cdancy/bitbucket/rest/features/CommentsApiLiveTest.java
cdancy: +1

src/main/java/com/cdancy/bitbucket/rest/options/UpdateTask.java
cdancy: Do we still need anchor? The API suggests we don't.

cdancy: If we get rid of anchor I have a way where we will be able to get rid of the `UpdateTask` altogether and have the user just pass along string/int properties which would simplify things but ONLY if we _can_ get rid of anchor.

README.md
```diff
@@ -179,8 +179,11 @@ Running integration tests can be done like so (requires Bitbucket instance):
 
     ./gradlew integTest
 
-Various [properties](https://github.com/cdancy/bitbucket-rest/tree/master/gradle.properties) exist for you to configure how the `integTest` task can be run should the defaults not suffice.
-	
+Various [properties](https://github.com/cdancy/bitbucket-rest/tree/master/gradle.properties) exist for you to configure
+how the `integTest` task can be run should the defaults not suffice. These properties can be overridden by specifying
+`-Pproperty=value` on the commandline. For example, to override the integTest credentials, `./gradlew
+-PtestBitbucketRestCredentials=user:pass integTest`.
+
```
cdancy: Looks good and thanks!

src/main/java/com/cdancy/bitbucket/rest/options/UpdateTask.java
ftclausen: After some pruning down of extra API details I got to the point where I see what you mean. So, in 007664f, I removed the `UpdateTask` class and just modified the TasksApi to accept a string representing the state. Followed up with modifying the tests to reflect this. 

I think this is a much nicer API to call although having the client code ensure it sends the correct strings is not ideal (but not sure what we can do about it since we're building on an existing API). Ideally this could just be a toggle method accepting a boolean for setting a task to opened or resolved.

ftclausen: Also, the reason I had made things more complicated with regard to request payloads is because I used Chrome developer tools to see what payloads were sent when using the Web UI. I _should_ have just read the fine docs to see what was really required versus extra stuff :)

src/main/java/com/cdancy/bitbucket/rest/features/TasksApi.java
```diff
@@ -52,6 +54,15 @@
...
     Task create(@BinderParam(BindToJsonPayload.class) CreateTask createTask);
+
+    @Named("tasks:update")
+    @Documentation({"https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-rest.html#idm8288367088"})
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Path("/{taskId}")
+    @Fallback(BitbucketFallbacks.TaskOnError.class)
+    @Payload("%7B \"state\": \"{state}\" %7D")
+    @PUT
+    Task update(@PathParam("taskId") int taskId, @PayloadParam("state") String state);
```
cdancy: This looks much cleaner. Very awesome.

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
```diff
@@ -121,15 +120,31 @@ public void testGetTaskOnError() {
...
+        assertThat(instanceNowResolved).isNotNull();
+        assertThat(instanceNowResolved.errors().isEmpty()).isTrue();
+        assertThat(instanceNowResolved.state()).isEqualTo("RESOLVED");
+
+        final Task instanceReopened = api().update(instanceNowResolved.id(),"OPEN");
+        assertThat(instanceReopened).isNotNull();
+        assertThat(instanceReopened.errors().isEmpty()).isTrue();
+        assertThat(instanceReopened.state()).isEqualTo("OPEN");
+    }
+
```
cdancy: Lets also add another test to check when things fail. Maybe pass in a bogus taskId and ensure that the `errors()` object gets properly populated.

cdancy:
Looks good. Couple more comments to address and I think we should be good to go.

src/main/java/com/cdancy/bitbucket/rest/features/TasksApi.java
ftclausen: 👍

src/test/java/com/cdancy/bitbucket/rest/features/TasksApiLiveTest.java
ftclausen: I have added an on failure test in 4ba5d22

ftclausen:
> Looks good. Couple more comments to address and I think we should be good to go.

I _think_ have addressed the outstanding items but, as you might expect, I am used to working in Bitbucket so I may have missed something. Let me know if that is the case.

Once everything is looking OK I'll cleanup the feature branch history.

cdancy:
@ftclausen everything LGTM. Going to run the integTests in a bit to confirm all works as expected but code-wise everything is good. As far as "cleaning up the branch and/or squashing" I've been not wanting to do that as of late. I've always like the idea, and have done so in my projects, but it removes the ability to be able to cherry-pick out and/or revert single commits. Does make the history look very clean but at the cost of being able to pin-point very specific commits. I would say leave things as-is and if all goes well with testing here I'll merge later today.

ftclausen:
> @ftclausen everything LGTM. Going to run the integTests in a bit to confirm all works as expected but code-wise everything is good. As far as "cleaning up the branch and/or squashing" I've been not wanting to do that as of late. I've always like the idea, and have done so in my projects, but it removes the ability to be able to cherry-pick out and/or revert single commits. Does make the history look very clean but at the cost of being able to pin-point very specific commits. I would say leave things as-is and if all goes well with testing here I'll merge later today.

No worries about rewriting branch history - I'm fine to leave it as is. My integTest runs have been OK on a development instance of Bitbucket server 5.16.3 (not tried 6 yet).

cdancy:
@ftclausen changes have been merged to master. I did have to revert your changes to `TestUtilities` as they were breaking for me. Otherwise everything else looks good.

cdancy:
Version `2.5.2` has been released.

ftclausen:
> @ftclausen changes have been merged to master. I did have to revert your changes to `TestUtilities` as they were breaking for me. Otherwise everything else looks good.

Great news, and no problem at all for reverting my changes. I'll figure out what was going on on my end when I get to the next contribution (fetching all tasks for a PR).

cdancy:
@ftclausen sounds good.

Issue #183, opened 2019-03-18
Implement `Users` endpoints


cdancy:
A new `UsersApi` should be created along with PR's for each endpoint.

https://docs.atlassian.com/bitbucket-server/rest/6.0.0/bitbucket-rest.html#idp362

ajammil:
Hello mr @cdancy,
I almost did the first endpoint, but I need feedback.
Also in the Mock test for the UserApi, I disabled the assertSent(..) function as it gave me errors, Could you tell me what I did wrong?
Thank you in advance.
https://github.com/ajammil/bitbucket-rest/tree/UserApi-Development

cdancy:
Feel free to open a pull request to get feedback earlier as well.

 I'll take a look at things in more depth shortly...

PR #189, opened 2019-03-28
UserApi Development Initialized

ajammil:
I did the first endpoint, but I need the feedback.
Also in the Mock test for the UserApi, I disabled the assertSent function as it gave me errors. Could you tell me what I did wrong?
Thank you in advance.

This PR is targeting issue #183 .

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/features/UserApi.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/features/UserApi.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.cdancy.bitbucket.rest.features;
+
+import com.cdancy.bitbucket.rest.annotations.Documentation;
+import com.cdancy.bitbucket.rest.domain.admin.UserPage;
+import com.cdancy.bitbucket.rest.fallbacks.BitbucketFallbacks;
+import com.cdancy.bitbucket.rest.filters.BitbucketAuthenticationFilter;
+import org.jclouds.javax.annotation.Nullable;
+import org.jclouds.rest.annotations.Fallback;
+import org.jclouds.rest.annotations.RequestFilters;
+
+import javax.inject.Named;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.MediaType;
+
+@Produces(MediaType.APPLICATION_JSON)
+@RequestFilters(BitbucketAuthenticationFilter.class)
+@Path("/rest/api/{jclouds.api-version}/user")
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+public interface UserApi {
+
+    @Named("user:get-user")
+    @Documentation({"https://docs.atlassian.com/bitbucket-server/rest/6.0.0/bitbucket-rest.html#idp362"})
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Path("/groups/more-members")
+    @Fallback(BitbucketFallbacks.UserPageOnError.class)
+    @GET
+    UserPage users(@Nullable @QueryParam("filter") String filter,
+                   @Nullable @QueryParam("group") String group,
+                   @Nullable @QueryParam("permission") String globalRole,
+                   @Nullable @QueryParam("permission.1") String projectRole,
+                   @Nullable @QueryParam("permission.1.projectKey") String projProjectKey,
+                   @Nullable @QueryParam("permission.1.projectId") String projProjectId,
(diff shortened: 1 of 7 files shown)


src/test/java/com/cdancy/bitbucket/rest/features/UserApiMockTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Map;
+
+import org.testng.annotations.Test;
+
+import com.cdancy.bitbucket.rest.BaseBitbucketMockTest;
+import com.cdancy.bitbucket.rest.BitbucketApi;
+//import com.cdancy.bitbucket.rest.BitbucketApiMetadata;
```
cdancy: Lets remove this if it's not needed.

src/test/java/com/cdancy/bitbucket/rest/features/UserApiMockTest.java
```diff
@@ -0,0 +1,94 @@
...
+                    null, null, null, null, null, 0, 2);
+            assertThat(up).isNotNull();
+            assertThat(up.errors()).isEmpty();
+            assertThat(up.size() == 1).isTrue();
+            assertThat(up.values().get(0).slug().equals("jcitizen")).isTrue();
+
+            final Map<String, ?> queryParams = ImmutableMap.of("filter", "jcitizen",
+                    "group","stash-users" , "permission", "ADMIN","start",0,"limit",2);
+            //assertSent(server, getMethod, restApiPath + BitbucketApiMetadata.API_VERSION
+            //        + "/users", queryParams);
```
cdancy: What is the failure you are seeing here?

src/main/java/com/cdancy/bitbucket/rest/features/UserApi.java
```diff
@@ -0,0 +1,61 @@
...
+    @GET
+    UserPage users(@Nullable @QueryParam("filter") String filter,
+                   @Nullable @QueryParam("group") String group,
+                   @Nullable @QueryParam("permission") String globalRole,
+                   @Nullable @QueryParam("permission.1") String projectRole,
+                   @Nullable @QueryParam("permission.1.projectKey") String projProjectKey,
+                   @Nullable @QueryParam("permission.1.projectId") String projProjectId,
+                   @Nullable @QueryParam("permission.2") String repoRole,
+                   @Nullable @QueryParam("permission.2.projectKey") String repoProjectKey,
+                   @Nullable @QueryParam("permission.3.projectId") String repoPrrojectId,
```
cdancy: So we could have N number of these which is going to make this a bit unruly and also limit folks if we only specify 3. We'll probably have to create an optional Binder which will take in a list of of these "permission objects" which we will then have to inject the values as query-params. Take a look [HERE](https://github.com/cdancy/bitbucket-rest/blob/master/src/main/java/com/cdancy/bitbucket/rest/binders/BindHookSettingsToPayload.java) and poke around the code base to see how we do this type of thing elsewhere. Let me know if you have any questions. Looks good so far.

src/test/java/com/cdancy/bitbucket/rest/features/UserApiMockTest.java
ajammil: alright, I will clean it up.

src/test/java/com/cdancy/bitbucket/rest/features/UserApiMockTest.java
ajammil: TravisCI is failing to pull, I want to show you the detailed errors from the build file. But I'm waiting for it to detect my commits.

cdancy: Do things fail locally when you run `./gradlew clean build mockTest`?

src/main/java/com/cdancy/bitbucket/rest/features/UserApi.java
ajammil: I'm sorry I misunderstood the documentation I thought permission was for global 1 was for project permissions and 2 for repo etc.. I didn't recognize they were the same thing but different filters.

I looked into what you told me, isn't that a way to bind some data to the payload? which is as much as I know should not be used with GET requests as their body is not processed? Please correct me if I misunderstood anything.

cdancy: Ignore the class name ... you can alter the request in general and append new query params/etc. Take a look at the below code where we do something like this:

https://github.com/cdancy/jenkins-rest/blob/master/src/main/java/com/cdancy/jenkins/rest/filters/ScrubNullFolderParam.java

src/test/java/com/cdancy/bitbucket/rest/features/UserApiMockTest.java
ajammil: Yes, they give me checkStyle error: Do not use Windows line endings

cdancy: Ok so not an issue with the test/assertion but more an issue with your gitconfig which you can play with to get this working as you'd expect.

ajammil: No, it doesn't even get to check assertions it fails from the beginning.

cdancy: Are you developing on windows or on a unix based machine? I do all development on unix so I'm not sure where that would be coming from. To fix things you _could_ just write the line out again :)

ajammil: I am using Ubuntu for my developments, I will try to figure out what's wrong and get back to you.

ajammil: I found out where the Mock errors were, also I removed that error from my checkStyle so it's easier to build locally, I didn't stage it tho.

src/main/java/com/cdancy/bitbucket/rest/features/UserApi.java
ajammil: I am sorry I have to postpone it to the next Friday, as am having my midterm exams this week and I got a lot of studying to do.

cdancy: @ajammil of course and take your time.

PR #191, opened 2019-04-16
RepositoryApi: Listing all repos regardless of project

### LINKED ISSUE #190: RepositoryApi: get all public/private repos together without looping projects
RepositoryApi#list throws exception when sending project as null. I know that Bitbucket Rest API supports listing all repos together without looping through all projects one by one.

I couldn't find anything about that, so would like to know if there is a planned roadmap about this or if it already exists, and if the exception thrown by RepositoryApi#list is thrown intentionally or if it is a bug.

One more note, Bitbucket supports creating repositories without associating with a project.

mronus:
- Update base path for RepositoryApi
- Create a new endpoint in RepositoryApi to list all repositories regardless of project

closes #190

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/features/RepositoryApi.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/features/RepositoryApi.java
@@ -51,14 +51,14 @@ import org.jclouds.rest.annotations.ResponseParser;
 
 @Produces(MediaType.APPLICATION_JSON)
 @RequestFilters(BitbucketAuthenticationFilter.class)
-@Path("/rest/api/{jclouds.api-version}/projects")
+@Path("/rest/api/{jclouds.api-version}")
 @SuppressWarnings("PMD.AvoidDuplicateLiterals")
 public interface RepositoryApi {
 
     @Named("repository:create")
     @Documentation({"https://developer.atlassian.com/static/rest/bitbucket-server/latest/bitbucket-rest.html#idm45888277587248"})
     @Consumes(MediaType.APPLICATION_JSON)
-    @Path("/{project}/repos")
+    @Path("/projects/{project}/repos")
     @Fallback(BitbucketFallbacks.RepositoryOnError.class)
     @POST
     Repository create(@PathParam("project") String project,
@@ -67,7 +67,7 @@ public interface RepositoryApi {
     @Named("repository:get")
     @Documentation({"https://developer.atlassian.com/static/rest/bitbucket-server/latest/bitbucket-rest.html#idm45888277593152"})
     @Consumes(MediaType.APPLICATION_JSON)
-    @Path("/{project}/repos/{repo}")
+    @Path("/projects/{project}/repos/{repo}")
     @Fallback(BitbucketFallbacks.RepositoryOnError.class)
     @GET
     Repository get(@PathParam("project") String project,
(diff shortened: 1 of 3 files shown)


src/main/java/com/cdancy/bitbucket/rest/features/RepositoryApi.java
```diff
@@ -98,17 +98,30 @@ RequestStatus delete(@PathParam("project") String project,
...
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Path("/repos")
+    @Fallback(BitbucketFallbacks.RepositoryPageOnError.class)
+    @GET
+    RepositoryPage listAll(@Nullable @QueryParam("name") String name,
+                           @Nullable @QueryParam("projectname") String projectName,
+                           @Nullable @QueryParam("permission") String permission,
+                           @Nullable @QueryParam("visibility") String visibility,
+                           @Nullable @QueryParam("start") Integer start,
+                           @Nullable @QueryParam("limit") Integer limit);
```
cdancy: Lets just rename this method `list` and remove the old method just above. 

Lets also put the `@Nullable @QueryParam("projectname") String projectName` as the first parameter here but change the last part to just `project` to keep inline with other naming conventions in this file.

cdancy: With the removal of the above method I'd like to keep all existing mock and integ tests to ensure they still work as expected. They will have to be re-worked slightly to account for the new parameters which will largely be null.

cdancy:
@mronus now I see where you are going with this and I like it. Couple comments to address but all-in-all LGTM.

src/main/java/com/cdancy/bitbucket/rest/features/RepositoryApi.java
mronus: I would agree with you if I look from project name perspective, but I think `list` and `listAll` are 2 different methods and we should keep them both. The method `list` takes project key as parameter and `listAll` takes project name as parameter. Both of them can be used for different use cases, but I can remove the old one if you don't agree.

cdancy: hmmm ... can you confirm if in this case `projectKey` and `projectName` can be used interchangeably? If so then lets do what I propose otherwise we'll stick with what you have.

cdancy: I'm starting to think that you mat be right the more I mull this is over. it is 2 separate endpoints and so maybe supporting both is more correct. Maybe we just change the name so that it's clearly intended to be different. Something like `listBy` to denote options are present. Thoughts?

mronus: I agree because `list` and `listAll` might be confusing together. Some options came to my mind:

`listByProject` & (`listAll` or `list`)
`listProjectRepositories` & (`listAll` or `list`)

WDYT?

cdancy: @mronus I actually don't like your suggestions nor my previous suggestion :) I think I'm just going to make a call here and say we stick with your original idea and use `listAll` as I don't have a better idea and that seems to get the point across. 

What I would like to change however is the naming and ordering of the parameters. Lets reverse the `name` and `projectname` params to match what we are doing elsewhere and rename to the below to keep the same standard going:
```
    RepositoryPage listAll(@Nullable @QueryParam("projectname") String project,
                           @Nullable @QueryParam("name") String repo
```

And my apologies for all the back and forth and changing my mind on things. I was hoping we could have 1 endpoint that does both things but in the end these clearly serve 2 separate purposes.

mronus: Np! It is good to see we came to a point :) However, I did a mistake and deleted my fork before this PR gets closed. So this PR remained orphan. I created a new one with latest changes we agreed on. Please check #192

cdancy:
Closing in favor of #192

PR #192, opened 2019-04-18
RepositoryApi: Listing all repos regardless of project ( copy of #191) 

### LINKED ISSUE #190: RepositoryApi: get all public/private repos together without looping projects

mronus:
- Update base path for RepositoryApi
- Create a new endpoint in RepositoryApi to list all repositories regardless of project

closes #190

### DIFF
--- a/src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiLiveTest.java
+++ b/src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiLiveTest.java
@@ -106,6 +106,25 @@ public class RepositoryApiLiveTest extends BaseBitbucketApiLiveTest {
     }
 
     @Test
+    public void testListAllRepositories() {
+        final RepositoryPage repositoryPage = api().listAll(null, null, null, null, 0, 100);
+
+        assertThat(repositoryPage).isNotNull();
+        assertThat(repositoryPage.errors()).isEmpty();
+        assertThat(repositoryPage.size()).isGreaterThan(0);
+
+        assertThat(repositoryPage.values()).isNotEmpty();
+        boolean found = false;
+        for (final Repository possibleRepo : repositoryPage.values()) {
+            if (possibleRepo.name().equals(repoKey)) {
+                found = true;
+                break;
+            }
+        }
+        assertThat(found).isTrue();
+    }
+
+    @Test
     public void testDeleteRepositoryNonExistent() {
         final String random = randomStringLettersOnly();
         final RequestStatus success = api().delete(projectKey, random);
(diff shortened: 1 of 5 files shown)


src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiLiveTest.java
```diff
@@ -105,6 +105,25 @@ public void testListRepositories() {
...
+        boolean found = false;
+        for (final Repository possibleRepo : repositoryPage.values()) {
+            if (possibleRepo.name().equals(repoKey)) {
+                found = true;
+                break;
+            }
+        }
+        assertThat(found).isTrue();
+    }
+
```
cdancy: Lets add a live test to exercise the failure case. I guess you _could_ put in say bogus values for things just to get bitbucket to return an error state. I do this for every endpoint so there should be multiple examples within this file.

src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiMockTest.java
```diff
@@ -325,6 +325,68 @@ public void testGetRepositoryListNonExistent() throws Exception {
...
+            assertThat(repositoryPage.limit()).isEqualTo(limit);
+            assertThat(repositoryPage.isLastPage()).isFalse();
+            assertThat(repositoryPage.nextPageStart()).isEqualTo(size);
+            assertThat(repositoryPage.values()).hasSize(size);
+            assertThat(repositoryPage.values()).hasOnlyElementsOfType(Repository.class);
+        } finally {
+            server.shutdown();
+        }
+    }
+
```
cdancy: Same thing here: lets add a live test to exercise the failure case. I guess you _could_ put in say bogus values for things just to get bitbucket to return an error state. I do this for every endpoint so there should be multiple examples within this file.

src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiLiveTest.java
mronus: Done. Please review the last commit when you have time.

src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiMockTest.java
mronus: Done. Please review the last commit when you have time.

cdancy:
Thanks for the PR @mronus . Kicking a point release shortly.

cdancy:
`v2.5.3` has been released with these changes.

PR #200, opened 2019-07-17
fix RepositoryApiLiveTest

### LINKED ISSUE #199: The RepositoryApiLiveTest does not work
The integrationtests fail on master.

### Expected Behavior

The integrationtests should not fail.

### Current Behavior

The RepositoryApiLiveTest fails.
To be more precise the methods testListAllRepositoriesByRepository and testListAllRepositories fail because the assertion `assertThat(foundRepos).isNotEmpty();` fails.

### Context

I want to open a PR and tried to write integrationtests for my new code.
I encountered this Issue while running the integrationtests for the first time.

`./gradlew integTest`

### Your Environment

Docker for Windows:
Docker version 18.09.2, build 6247962
Atlassian Bitbucket v6.4.0

hesch:
The tests now work even if there is only one page of results.

closes #199

### DIFF
--- a/src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiLiveTest.java
+++ b/src/test/java/com/cdancy/bitbucket/rest/features/RepositoryApiLiveTest.java
@@ -114,7 +114,8 @@ public class RepositoryApiLiveTest extends BaseBitbucketApiLiveTest {
         int start = 0;
         final int limit = 100;
         RepositoryPage repositoryPage;
-        while (!(repositoryPage = api().listAll(null, null, null, null, start, limit)).isLastPage()) {
+        do {
+            repositoryPage = api().listAll(null, null, null, null, start, limit);
             start += limit;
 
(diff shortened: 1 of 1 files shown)


cdancy:
@hesch this has never failed before and I don't see how this fix changes things to all of a sudden not fail. The code in this PR is fine but the assertion comes after the `while` loop, and whether using your impl and or the one that currently exists, should still have the same results.

hesch:
If the first page received has the isLastPage attribute set to true, the condition in the while loop will be false and the code inside will never be executed. This means there will be no items added to the foundRepos list and the assertion evaluates to false.

My implementation always adds all the items from the first page and then checks whether we are on the last page or not. This way the items from the last page end up in the list and the assertion evaluated to true.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

cdancy:
@hesch thanks for the fix and looking forward to the impending PR you're putting together!

Issue #201, opened 2019-09-25
MVN Repository files not found


AlyIbrahim:
### Expected Behavior
Adding the maven dependency to my project and running 
mvn clean install should end up pulling the pom and jar file from maven repositories.

### Current Behavior
```Downloading from central: https://repo.maven.apache.org/maven2/com/cdancy/bitbucket-rest/2.5.3/bitbucket-rest-2.5.3.pom
[WARNING] The POM for com.cdancy:bitbucket-rest:jar:2.5.3 is missing, no dependency information available
Downloading from central: https://repo.maven.apache.org/maven2/com/cdancy/bitbucket-rest/2.5.3/bitbucket-rest-2.5.3.jar
```
And if you follow the link to the jar file, I am getting `404 Not Found`

### Context
I am trying to add bitbucket-rest as a dependency to my project.
I added the following dependency in my pom.xml
```
<dependency>
    <groupId>com.cdancy</groupId>
    <artifactId>bitbucket-rest</artifactId>
    <version>2.5.3</version>
</dependency>
```

### Steps to Reproduce (for bugs)
Create a java maven project
Add this dependency to your pom.xml file:
```
<dependency>
    <groupId>com.cdancy</groupId>
    <artifactId>bitbucket-rest</artifactId>
    <version>2.5.3</version>
</dependency>
```

run `mvn clean install`

### Your Environment
Linux RHEL 7.6
JDK 1.8.0_201
Maven 3.6.0

cdancy:
I actually don't publish to maven and ONLY publish to Bintray. I _assume_ things are getting propagated to maven maybe by Bintray but I never wired that up. Can you try pulling from Bintray and see if that resolves your problem?

AlyIbrahim:
Yep, finally it worked ..
It took me sometime to make it to pull from bintray, maybe it's me having the first time with bintray, but I suggest, we can add this info to the README to make it easier.
I can send a pull request to this ..
Actually it turned me away from few weeks back when I was not able to pull from Maven, but I like the work been done here and give it another try and finally end up opening this issue, but others can miss the chance to use this if they don't have the quick and easy way to make it work.
Thanks Chris

cdancy:
@AlyIbrahim yeah feel free to send in a PR to add to the documentation. I do note that you can resolve things through `jcenter` as that's largely interchangeable with `bintray` assuming folks know this. I've amended the line on the README.

AlyIbrahim:
Got it ..
[...]

PR #202, opened 2019-10-01
Maven bintray readme

AlyIbrahim:
**This PR is based on issue :** [MVN Repository files not found #201 ](https://github.com/cdancy/bitbucket-rest/issues/201)
The PR is adding to the steps of adding bitbucket-rest as a dependency to your maven project and providing 2 different ways to include bintray repo either in settings.xml or within your project pom.xml.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Can be sourced from jcenter like so:
 	
 ### Maven repository Configuration:
 
-* Download a sample settings.xml file that has configured profile to pull from the correct maven repository using this link: [settings.xml](https://bintray.com/repo/downloadMavenRepoSettingsFile/downloadSettings?repoPath=%2Fcdancy%2Fjava-libraries)
+* Download a sample settings.xml file that has configured profile to pull from the correct maven repository using this link: [settings.xml](https://bintray.com/repo/downloadMavenRepoSettingsFile/downloadSettings?repoPath=%2Fbintray%2Fjcenter)
 
 **OR**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,6 +39,27 @@ Can be sourced from jcenter like so:
       <classifier>sources|tests|javadoc|all</classifier> (Optional)
     </dependency>
 	
+**Note**: You need to have bintray maven repository configured to be able to pull bitbucket-rest dependency.
```
cdancy: Lets remove this line as that's already made clear further up above.

README.md
```diff
@@ -39,6 +39,27 @@ Can be sourced from jcenter like so:
...
+      <repository>
+        <snapshots>
+          <enabled>false</enabled>
+        </snapshots>
+        <id>bintray-cdancy-java-libraries</id>
+        <name>bintray</name>
+        <url>https://dl.bintray.com/cdancy/java-libraries</url>
+      </repository>
+    </repositories>
+```
```
cdancy: Would it be better to just do something like the below and link to it?

https://stackoverflow.com/questions/44265547/how-to-properly-specify-jcenter-repository-in-maven-config

README.md
AlyIbrahim: Agree, removed already

README.md
AlyIbrahim: @cdancy 
I can do this to the embeded one "second option"
But for the first option, settings.xml, I find it easier to provide the link and download the settings.xml instead of adding the text and making README very big.
I can check if jcenter is providing a link to a sample settings.xml, or else, we can add it in the repo and link to it
What do you think?

AlyIbrahim: @cdancy OK, I find a link to settings.xml for jcenter file, modified both options to use jcenter

AlyIbrahim:
Merge?

cdancy:
I swear I did this on my phone :) Merging now!

PR #205, opened 2019-10-23
Implement insights API

hesch:
I implemented the insights API. The endpoints are all together in this PR, because some of them depend on eachother. I hope that is not a problem.


cdancy:
@hesch this all LGTM. I'll give things a go to make sure everything passes.

cdancy:
All live tests pass as expected. Thanks for the PR @hesch!!! Kicking a new minor release now `2.6.0`

PR #206, opened 2019-10-24
fix checkstyle & pmd errors

hesch:


cdancy:
hmmm ... these should have been picked up during the travis run. Not sure why that did not pop. In either event thanks for the fix!

PR #207, opened 2019-10-24
udpate dependencies

hesch:
I updated some dependencies and the gradle version. The projects now also works with Java 12.


cdancy:
We'll take the updates! Thanks again @hesch !

PR #209, opened 2019-12-18
use correct data formats in InsightReportData value

### LINKED ISSUE #208: Code Insight API ReportData can have values of different types
### Expected Behavior
Adding ReportData of DataTypes other than TEXT should work

### Current Behavior
Only adding ReportData of DataType TEXT works, because other types need to be encoded as described here:
https://docs.atlassian.com/bitbucket-server/rest/6.9.0/bitbucket-code-insights-rest.html#idp9
But currently every type is encoded as a string.

### Context
I was trying to add a link to a report.

```
val data = InsightReportData.create("Title", InsightReportData.DataType.LINK, "https://example.com/insightReportData");
val report = CreateInsightReport.create("", "https://example.com/buildUrl", null, CreateInsightReport.RESULT.PASS, "", "", List.of(data));
insightsApi.createReport("project", "repo", "commit", "key", report);
```

### Your Environment
Java 11
Jenkins CI

So I already fixed this, but then read that I should create an issue first.
Just tell me if I should open a PR.

succcubbus:
Adding Code Insight ReportData of DataTypes other than TEXT should now work

closes #208 

During live testing I got a few `*** Please tell me who you are.` messages, and I don't know how to fix these.
So the live tests could be broken.

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/domain/insights/InsightReportData.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/domain/insights/InsightReportData.java
@@ -20,6 +20,11 @@ package com.cdancy.bitbucket.rest.domain.insights;
 import com.google.auto.value.AutoValue;
 import org.jclouds.json.SerializedNames;
 
+import java.time.Duration;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+
 @AutoValue
 public abstract class InsightReportData {
     public enum DataType {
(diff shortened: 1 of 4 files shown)


succcubbus:
The same integration tests are failing for me on master, so it's probably something about my local bitbucket instance.

cdancy:
@succcubbus got it working here locally. Will merge this in, apply fix, and kick a new point release. Thanks for the PR!

cdancy:
FWIW: had to massage the test to include less than or equal to 6 items in the `data` list as my bitbucket instance was complaining I could not send in a list with more than that. Not sure if that's a config someone can increase or a hard default.

cdancy:
Version `2.6.2` released with this fix. Thanks again!

succcubbus:
whoa that was fast
I expected to wait a bit for the next release
awesome

PR #240, opened 2020-10-02
Changed parameter name of query parameter "at" to "branchOrTag"

### LINKED ISSUE #230: FileApi.listLines: Wrong naming of parameter
In the FileApi, there is the function listLines, which gives plenty of parameters.
One of the parameters based on the Bitbucket documentation is "at", in your project it's "commitHash".

Unfortunately this is very confusing, as it is not referencing to a specific commit hash, but to a branch e.g.

So calling the API with `?at=refs/heads/master` or `?at=refs/tags/%myTag%` will give a valid result.
Putting a commitHash instead would produce an error message:
````
{
    "errors": [
        {
            "context": null,
            "exceptionName": "com.atlassian.bitbucket.content.NoSuchPathException",
            "message": "The path \"xxx\" does not exist at revision \"9aa913957f1...\""
        }
    ]
}
````

### Expected Behavior
To avoid misunderstanding, the parameter should be named like in Bitbucket documentation as there is at least some documentation there

### Current Behavior
The parameter `commitHash` is named wrongly and should be renamed

### Context
It is just annoying and misleading if you deal with that

### Your Environment
Version of bitbucket-rest: 2.6.3
Version of Bitbucket: 7.4.0

l0wlik34G6:
In the FileApi, there is the function listLines, which gives plenty of parameters.
One of the parameters based on the Bitbucket documentation is "at", in your project it's "commitHash".

Unfortunately this is very confusing, as it is not referencing to a specific commit hash, but to a branch e.g.

So calling the API with `?at=refs/heads/master` or `?at=refs/tags/%myTag%` will give a valid result.
Putting a commitHash instead would produce an error message:
````
{
    "errors": [
        {
            "context": null,
            "exceptionName": "com.atlassian.bitbucket.content.NoSuchPathException",
            "message": "The path \"xxx\" does not exist at revision \"9aa913957f1...\""
        }
    ]
}
````

### Expected Behavior
To avoid misunderstanding, the parameter should be named like in Bitbucket documentation as there is at least some documentation there

### Current Behavior
The parameter `commitHash` is named wrongly and should be renamed

### Context
It is just annoying and misleading if you deal with that

### Your Environment
Version of bitbucket-rest: 2.6.3
Version of Bitbucket: 7.4.0

----

As `at` is not very saying imho and `commitHash` or `commitIdOrRef` are wrong, I would suggest to change it to `branchOrTag`.

Closes #230

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/features/FileApi.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/features/FileApi.java
@@ -59,8 +59,8 @@ public interface FileApi {
     RawContent raw(@PathParam("project") String project,
                 @PathParam("repo") String repo,
                 @PathParam("filePath") String filePath,
-                @Nullable @QueryParam("at") String commitHash);
-    
+                @Nullable @QueryParam("at") String branchOrTag);
+
     @Named("file:list-lines")
     @Documentation({"https://developer.atlassian.com/static/rest/bitbucket-server/5.1.0/bitbucket-rest.html#idm45588158357840"})
     @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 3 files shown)


cdancy (approved):
LGTM and thanks for the cleanup. Completely agree and this makes things more obvious at least when using an IDE.

cdancy:
@l0wlik34G6 let me know if you want a new patch release. Will take 10 seconds to do.

Issue #241, opened 2020-10-17
Feature: Search API


thlaegler:
**This is a Feature Request**

### Expected Behavior
This Bitbucket REST API Client implementation should provide the `Search API` Feature

### Current Behavior
Not implemented

### Context
It's not officially documented - at least I couldn't find something? But there is a Search API that gets called from the Bitbucket Server Frontend when doing a Search (e.g. across all repos). So, the API is not like the others under `/rest/api/...` but under `/rest/search/...`.
There is a workaround to search via REST by calling Elasticsearch directly https://community.atlassian.com/t5/Bitbucket-questions/Rest-API-to-search-a-string-in-all-the-repositories-in-Bitbucket/qaq-p/900814.
But in my case I don't have access to ES or port `9200` on the Bitbucket Server.

Working CURL example; tested on our Bitbucket Server v7.3.1:
```
curl --location --request POST 'https://<your-bitbucket-server>/rest/search/1.0/search' \
--header 'Accept: application/json' \
--header 'Authorization: Basic <base64-of:username:personal-access-token>' \
--header 'Content-Type: application/json' \
--data-raw '{
  "query": "<my-search-phrase>",
  "entities": {
    "code": {}
  },
  "limits": {
    "primary": 25,
    "secondary": 10
  }
}'
```

### Your Environment
Tested on Atlassian Bitbucket Server `7.3.1` with API-versions `1.0` and `latest`.

thlaegler:
Documentation of Search feature from UI perspective: https://confluence.atlassian.com/bitbucketserver/search-for-code-in-bitbucket-server-814204781.html. That goes into the `query` element as search phrase.

PR #242, opened 2020-10-17
Bitbucket Server: Search API (/rest/search/{version}/search)

thlaegler:
Implementation of Feature `Search API`: #241


cdancy (approved):
@thlaegler I'm fine with this. Anything else you want to add or do here?

thlaegler:
I'm fine with it. Would be good to test it against other Bitbucket versions. Maybe even Bitbucket Cloud.

cdancy:
@thlaegler what did you test against? We have v7.4.0 here that I can run against.

Also ... would it make sense to allow for passing in a raw string to the "SearchRequest" object on top of the map and json versions we have now? I could see that being useful for users who want to quickly build something out with a string.

thlaegler:
I tested with v7.3.1
Good idea with quick search. Actually, paging could also be implemented. But it looks like the paging works a bit different than the other paging-APIs so I didn't implemented it

cdancy:
@thlaegler sorry for the delay. There were checkstyle issues when I ran locally and didn't have the time to address them until now. `2.7.1` is now released and should be good to go.

PR #254, opened 2021-01-15
Use user.name() instead of user.slug() to avoid case issues

### LINKED ISSUE #253: integTest: TestUtilities class should use user.name() instead of user.slug() to avoid case issues
The TestUtilities class is using the equivalent of the `User` class' "name" field when retrieving the name from `auth.authValue()`. If a username has an upper case letter (e.g. JoeSmith) then the tests will fail because the `user.slug()` is all lower case (joesmith). Using user.name seems to fix this - PR forthcoming.

ftclausen:
closes #253 

This change makes the username comparison in TestUtilities consistent with how the username is presented when getting it from authentication info (`auth.authValue()`). Using `user.slug()` means the username is all lower case causing integ tests to fail if the user had put in a username like `JoeSmith` because the comparison ends up being on `joesmith != JoeSmith`. 

By using `user.name()` we get back the same name (in terms of case) as it seems to be stored in the auth info `auth.authValue()`. Login does not appear to be case sensitive though - so an alternative approach is just to normalise all username strings to lowercase and compare that.

### DIFF
--- a/src/test/java/com/cdancy/bitbucket/rest/TestUtilities.java
+++ b/src/test/java/com/cdancy/bitbucket/rest/TestUtilities.java
@@ -79,7 +79,7 @@ public class TestUtilities extends BitbucketUtils {
                 assertThat(userPage).isNotNull();
                 assertThat(userPage.size() > 0).isTrue();
                 for (final User user : userPage.values()) {
-                    if (username.equals(user.slug())) {
+                    if (username.equals(user.name())) {
                         defaultUser = user;
                         break;
                     }


cdancy:
@ftclausen hey how are things going? Code looks fine to me. Lets take it in and thanks for the PR!

ftclausen:
All well thanks! We're upgrading from self-hosted Bitbucket 5.16.x to 7.6.x so seeing how the bitbucket-rest integ tests go against that version.

Issue #255, opened 2021-01-15
Label api support - Filter by name


shinusuresh:
Provide API support to label filter by name endpoint
https://docs.atlassian.com/bitbucket-server/rest/6.0.0/bitbucket-rest.html#idp88

PR #256, opened 2021-01-15
Support to labels endpoint - filter by labelname

shinusuresh:
PR for #255

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/features/LabelsApi.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/features/LabelsApi.java
@@ -18,6 +18,7 @@
 package com.cdancy.bitbucket.rest.features;
 
 import com.cdancy.bitbucket.rest.annotations.Documentation;
+import com.cdancy.bitbucket.rest.domain.labels.Labels;
 import com.cdancy.bitbucket.rest.domain.labels.LabelsPage;
 import com.cdancy.bitbucket.rest.fallbacks.BitbucketFallbacks;
 import com.cdancy.bitbucket.rest.filters.BitbucketAuthenticationFilter;
@@ -29,6 +30,7 @@ import javax.inject.Named;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.MediaType;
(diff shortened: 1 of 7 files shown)


cdancy:
@shinusuresh looks like a failing mock test?

shinusuresh:
My bad @cdancy. Missed to add mock JSON file.

shinusuresh:
@cdancy Can you have a look into this PR, please

src/main/java/com/cdancy/bitbucket/rest/features/LabelsApi.java
```diff
@@ -45,4 +47,12 @@
...
     @GET
     LabelsPage list(@Nullable @QueryParam("prefix") String prefix);
+
+    @Named("labels:get-by-name")
+    @Documentation({"https://docs.atlassian.com/bitbucket-server/rest/6.0.0/bitbucket-rest.html#idp88"})
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Path("/api/{jclouds.api-version}/labels/{labelName}")
+    @Fallback(BitbucketFallbacks.LabelsByNameOnError.class)
+    @GET
+    Labels getLabelByName(@Nullable @PathParam("labelName") String labelName);
```
cdancy, line 57: @shinusuresh should we allow `@Nullable` here? What happens if a user passes in null?

cdancy:
@shinusuresh one comment to address. Are there any labels we could query for that would be present by default for integration tests?

src/main/java/com/cdancy/bitbucket/rest/features/LabelsApi.java
shinusuresh: @cdancy Its, not mandatory as per the spec. 
I think it needs to be handled. I am updating PR

shinusuresh:
> @shinusuresh one comment to address. Are there any labels we could query for that would be present by default for integration tests?
It's a little tricky. Labels are being set at the project level, and these API's are not tied to any project. 
I think when we implement https://docs.atlassian.com/bitbucket-server/rest/6.0.0/bitbucket-rest.html#idp227, these test can be amended to create a label first and then assert.
What you think?

src/main/java/com/cdancy/bitbucket/rest/features/LabelsApi.java
shinusuresh: @cdancy Is it a `NullPointerException` expected if the user passes in null?

cdancy: Can we just remove the @Nullable annotation and force users to pass in the parameter? Feels odd to have it here.

Also ... can we rename the `Labels` object to just `Label` as that would better represent what is being returned. Pluralizing it gives the impression that potentially multiple labels are getting returned.

cdancy:
> amended to create a label first and then assert.
> What you think @cdancy ?

That works for me.

cdancy (approved):
LGTM

cdancy:
@shinusuresh assuming no last minute issues all looks good to me and thanks for addressing the comments. I can merge once CI testing has passed. Did you want a new release with these changes or did you want to wait a bit and submit some more code first? I'm fine either way and will leave it up to you.

shinusuresh:
> Did you want a new release with these changes

@cdancy thanks for checking. Not needed now. Will release it, once Labels API is finished.

PR #446, opened 2025-01-18
Fix build and bump dependencies

hpoettker:
Hi! Thanks for this great project. It has been serving me well for quite some years by now.

### Apache Jclouds 2.6.0

Activity on Apache Jclouds has unfortunately diminished but they have released 2.6.0, which contains updates to the latest dependencies. In particular, it now works with recent versions of Gson and uses Jakarta dependencies instead of Java EE.

The version of the Jakarta XML bindings has only been raised to 2.3.3, which means that the imports for the bindings still use `javax`. But I guess, we'll have to live with that.

### Changes in this PR

The version of Apache Jclouds has already been bumped but the imports still use the `javax` namespace, wherefore builds of `master` currently fail. Furthermore, there are some incompatibilities in the build and test dependencies.

To get the build working again, I've

- replaced the `javax` imports with `jakarta` imports
- upgraded Gson to the latest version
- upgraded the Jakarta EE libraries to those versions used by Jclouds 2.6.0
- upgraded Gradle to the latest release of 7.x
- upgraded the Gradle shadow plugin
- fixed one style issue that PMD was complaining about in `BitbucketFallbacks`

I've also bumped some more libraries to make dependabot happy.

The build is now green, the mock tests are successful, I've successfully tested the standalone and the shaded jar in my application.


hpoettker:
Apache Jclouds is preparing a release of 2.7.0: https://lists.apache.org/thread/9htjc47nwn0cjjz2pkod46thygowt92k

If I understand correctly, this will also upgrade the XML bindings to the Jakarta namespace.

I'll update the PR when 2.7.0 is released.

cdancy:
Hey ... just seeing this now and thanks for the PR! Let me know when they release a new version and we can get that in.

hpoettker:
Apache jclouds 2.7.0 has been released.

I've bumped its version and upgraded the dependency on `jakarta.xml.bind-api` to 4.0.2. So there should be no more dependency on something with the javax namespaces that have been migrated to Jakarta.

hpoettker:
@cdancy If you find the time, it would we great if we could get this done and released.

The library itselfs works great but that the latest release depends on older versions of Java EE and Gson creates more and more hoops to jump through in applications with higher base lines for the two.

And I've also pushed a bugfix that would be great to have released: #448

cdancy:
@hpoettker apologies as I didn't realize you finished things. Let me have a look now ...

cdancy:
@hpoettker have you build locally and tested that things work? Have you tried the `all` artifact as well to ensure it builds successfully and works "as is"?

hpoettker:
I've built locally with JDK 17, run the mock tests, disabled the jar-signing, and then published to Maven Local.

I've then used both the standard and the all-jar successfully with my applications. The all-jar works fine, but it still has the issue that it pulls in the unshaded transitive dependencies although they are shaded into the jar. I didn't address this in this PR but it would make sense to address it in a follow-up PR.

cdancy:
>  I didn't address this in this PR but it would make sense to address it in a follow-up PR.

@hpoettker yeah I never got around to fixing that but when we use it here internally we just set `trasivite dependencies to false` and that seems to work.

Ok ... this looks good. I'm going to run some tests here to make sure all works OK and if so I'll do a release later today.

hpoettker:
I had a look how the pulling in of transitive dependencies by the shaded jar could be fixed. But as long as the jars are published under the same artifact identifier, you can only have one pom file, and it's the pom file that lists the dependencies. So this isn't really possible.

While I had a look at this, I noted that the relocations in the shaded jar are incomplete. They were also incomplete in the 3.1.1 release, where e.g. `com.sun`, `javax.activation`, and `org.checkerframework` aren't relocated.

I assume everything should actually be relocated. I'll push the required change.

cdancy:
@hpoettker merged this PR. Looking at your other PR and some others as well and getting all of those in for a new release.

PR #448, opened 2025-02-07
Fix NPE in error response parsing

### LINKED ISSUE #447: NullPointerException when parsing error responses
After upgrading a Bitbucket Data Center instance from 8.x to 9.4.1, we have observed a change in the response body for errors.

For example when sending a `CreatePullRequest` with the `PullRequestApi` to merge `feature` into `main` where the two branches are identical, we get a 409 Conflict return status with a response body like

```
{
  "errors": [
    {
      "message": "Branch \"main\" is already up-to-date with branch \"feature\" in repository \"my-repo\".",
      "exceptionName": "com.atlassian.bitbucket.pull.EmptyPullRequestException"
    }
  ]
}
```

The difference to before is that the field `context` is not contained in the error element. The current implementation allows the field be explicitly set to `null` but when the field is missing, the following error is returned by the `PullRequestApi`:

```
{
  "message": "Failed to parse output: message=Cannot invoke \"com.google.gson.JsonElement.isJsonNull()\" because \"context\" is null"
  "exceptionName": "java.lang.NullPointerException"
}
```

hpoettker:
Resolves #447

I've successfully tested the commit based on the branch of #446. The mock tests are successful and it works in my application against a real-world Bitbucket instance.

This PR here is against the `master` branch, which currently doesn't build. I'll rebase after #446 has been merged.

### DIFF
--- a/src/main/java/com/cdancy/bitbucket/rest/fallbacks/BitbucketFallbacks.java
+++ b/src/main/java/com/cdancy/bitbucket/rest/fallbacks/BitbucketFallbacks.java
@@ -911,7 +911,7 @@ public final class BitbucketFallbacks {
 
         final List<Veto> vetos = Lists.newArrayList();
         final JsonElement possibleVetoesArray = obj.get("vetoes");
-        if (possibleVetoesArray != null && !possibleVetoesArray.isJsonNull()) {
+        if (isNotJsonNull(possibleVetoesArray)) {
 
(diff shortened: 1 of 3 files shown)


hpoettker:
@cdancy I'd be happy to contribute more fixes and additions. But this really only makes sense when there are also releases to Maven Central. Otherwise, maintaining a private fork is really simpler.

I don't feel strongly about #452 or any other open issue or PR. But a release of the current master would be very much appreciated.

